PR #46, opened 2020-03-04
Add keybindings for monitor up/down

herder:
This adds Ctrl-Super-Shift-<up/down> keybindings to move windows up and down monitors.
I have my screens setup so that my laptop is below my external screen, and without this change
there is no way to move windows between my screens.

The keybindings are not perfectly consistent imho, but I'll add an issue with a suggestion about that.

Thanks for a great plugin! This is the only thing I really missed in Pop, and I'm very happy that you did too :)


herder:
Never mind - this seems to work in Super+Enter using the direction keys. Thanks!

Closing.

PR #63, opened 2020-03-12
Forest replaces AutoTiler into panel_settings.ts 

senaki:
Make listen command reports an error with auto_tiler js module missing into panel_settings.js.
Checking the history of modifications into src *.ts file reveals that  auto_tiler module is replaced by forest.
Doing the same into panel_settings.ts removes the error and the building works correctly.


mmstick:
Sorry. Didn't see this PR. I've fixed it

senaki:
> Sorry. Didn't see this PR. I've fixed it

No problem :)

PR #65, opened 2020-03-12
Fix #64 user theme styles being overridden

### LINKED ISSUE #64: Pop-shell forces Dash To Dock border to be rounded
With pop-shell enabled, the dash to dock border radius is forced to be a certain value, regardless of what the dash to dock "force straight corner" setting is set to.

This can be seen here:
[pop-shell enabled](https://user-images.githubusercontent.com/3020843/76548704-12faa480-6487-11ea-815a-4e3e37c3270d.png)
[pop-shell disabled](https://user-images.githubusercontent.com/3020843/76548707-142bd180-6487-11ea-971c-1a228f9b8107.png)

System:
Archlinux on linux 4.19-lts
Gnome-shell 3.34.4
Pop-shell git latest (r233.6b0cd23-1)

Benguin:
Fixes the issue in #64. I'm not experienced with gnome-shell extension development so I'm not sure if this is the correct fix, but it did work for me :)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1009,7 +1009,7 @@ function load_theme() {
         let theme = new St.Theme({
             application_stylesheet: Gio.File.new_for_path(Me.path + "/stylesheet.css"),
             theme_stylesheet: getThemeStylesheet(),
-            default_stylesheet: _defaultCssStylesheet,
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks!

PR #76, opened 2020-03-15
Replace get_key_code() with get_key_symbol()

jfkimmes:
The hardware key_code might not be identical to the symbol the user
intends to press (e.g in case of key remapping).


mmstick:
Thanks!

PR #86, opened 2020-03-18
Set rebuild.sh to use bash

MiguelMoll:
Not sure if this was left out deliberately but it helps when sh or bash isn't the active shell (e.g. fish). Open to alternatives (/usr/bin/sh).

Locally I'm running `bash ./rebuild/sh`.


MiguelMoll:
Docs state to use `sh ./rebuild.sh`. That should suffice.

Issue #138, opened 2020-04-02
Create TESTING.md and add branch protection


jackpot51:
This will be needed before the 20.04 release.

PR #173, opened 2020-04-15
Add ISSUE_TEMPLATE

brs17:
Adds default issue template

EDIT: This has nothing to do with https://github.com/pop-os/shell/issues/138


WatchMkr:
This is helpful. We need TESTING.md as well before release. We need to test each bugfix release after the initial release.

This can be removed from the template ```**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):**```

brs17:
> This can be removed from the template `` **Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):** ``

Why would we want to remove this if the first question we will almost always ask, is if the user is running the latest version of the shell?

WatchMkr:
You're right. I didn't see the version part of the message.

PR #177, opened 2020-04-15
Copy change and reduce launcher size

WatchMkr:


WatchMkr:
Can’t change show active hint. It’s in marketing material. Will open a new pr to reduce the launcher size.

PR #235, opened 2020-04-29
Fix exception for Firefox

jackpot51:
Fixes https://github.com/pop-os/beta/issues/2

### DIFF
--- a/src/window.ts
+++ b/src/window.ts
@@ -118,7 +118,7 @@ export class ShellWindow {
     }
 
     ignore_decoration(ext: Ext): any {
-        return this.name(ext) === 'Firefox';
+        return this.name(ext) === 'Firefox Web Browser';
     }
 
     may_decorate(): boolean {


jeroen7s:
not sure if hardcoding a specific application name is the best solution for this.
Tor browser has the same issue btw, maybe some other applications that i'm not yet aware of.

mmstick:
We can use an array as we're already doing for blacklisting certain windows from tiling that violate X11 window manager standards (applications not hinting the window manager that they are dialogs, utilities, etc.). Firefox doesn't tell us that it's doing its own brand on client-side rendering, so it needs some special case handling. We can add Tor too. What is the window name?

jeroen7s:
Window name for Tor is "Tor Browser"

jeroen7s:
Though I'm not sure if messing with window borders is nessecary for tiling? 
Shouldn't the default be "don't touch window borders", instead of forcing all borders either on or of?
Seems like the window name fix breaks with non-english languages. https://github.com/pop-os/beta/issues/2#issuecomment-622781464

PR #242, opened 2020-05-01
fix: Fullscreening a window will now actually fullscreen it

mmstick:
Without this, a fullscreening a window — such as when clicking the fullscreen
button in YouTube — causes the fullscreen'd window to be tiled. This is
caused by only checking for the maximization state on a size change,
when we should check both the maximization and fullscreen state.


laichiaheng:
Is the YouTube fullscreen caused by this?

PR #243, opened 2020-05-01
fix: Reset fork ratios to 1/2 on attaching a window to a window

mmstick:
Now correctly resets a fork so that a newly-tiled window attains half
the area of the fork.


mmstick:
Changes now in `dev_focal` branch

PR #251, opened 2020-05-02
Correct case for hotkeys

kalenpw:
Just fixed the case on some hotkeys since I initially tried `super` + `shift` + `o`

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ The launcher is summoned with `Super` + `/`. The search list displays matching w
 By default, the launcher searches windows and applications. However, you can designate a special launch mode using one of the supported prefixes:
 
 - `:`: Execute a command in `sh`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -143,7 +143,7 @@ The launcher is summoned with `Super` + `/`. The search list displays matching w
 By default, the launcher searches windows and applications. However, you can designate a special launch mode using one of the supported prefixes:
 
 - `:`: Execute a command in `sh`
-- `t:`: Execute a command in `sh` in a terminal
```
brs17, line 146: `t` does need to be followed by a `:` for this to work so I think it is correct as it is currently listed.

kalenpw:
That has been fixed.

WatchMkr:
We use capitals to match with GNOME shortcuts.

PR #264, opened 2020-05-03
Expand app search to include snaps and /usr/local

scaryrawr:
- Add /usr/local/share/applications to launcher search path
- Add snap applications to launcher search path


scaryrawr:
Fixes #265

Issue #227, opened 2020-04-27
Launcher window focus dissappears after unlocking screen


nicolae-stroncea:
**Distribution (run `cat /etc/os-release`):**
 
Fedora 32, Gnome 3.36, X11
 
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):**

586655441b3c281a8c873e36c2e3997b75b745bf, but occurs due to this commit 4d85c31f9fc85ef2625fd09f0e318607bca12dcf. 

**Issue/Bug Description:**

After unlocking screen, navigating through windows in launcher does not apply the grey background highlight to any window in the launcher.

Before 4d85c31f9fc85ef2625fd09f0e318607bca12dcf

![image](https://user-images.githubusercontent.com/39338488/80410608-dd741400-8898-11ea-8ac5-718bc8212c3b.png)

After 4d85c31f9fc85ef2625fd09f0e318607bca12dcf

![image](https://user-images.githubusercontent.com/39338488/80410449-9423c480-8898-11ea-91c3-35f43bcaae43.png)

**Steps to reproduce (if you know):**

1. Lock Screen
2. Unlock screen
3. `Super /`

PR #270, opened 2020-05-04
Load pop stylesheet into existing user theme

scaryrawr:
When using `loadTheme` it merges the default GNOME theme with the style sheet passed into `setThemeStylesheet`. This change instead uses the existing theme and loads the pop style sheet on top of it (if a previous theme exists).

Fixes #231

Issue #216, opened 2020-04-24
Active hint stops functioning after unlocking the screen


nicolae-stroncea:
**Distribution (run `cat /etc/os-release`):**
 
 Fedora 32, Gnome 3.36, Xorg
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):**

4d85c31f9fc85ef2625fd09f0e318607bca12dcf

**Issue/Bug Description:**

Haven't been able to reproduce what triggers it, but after some time, active hint always disappears.

nicolae-stroncea:
Found a way to reproduce it: active hint disappears after you lock and unlock the screen.

nicolae-stroncea:
4d85c31 seems to cause it. The commit before it works fine.

calumwhite:
Distribution: Pop Os 20.04 LTS updated from Pop Os 19.04

Can confirm the same behaviour. Active hints work on a fresh login, however disappear after the screen has been locked.

flolivaud:
I can confirm the same issue. Fresh install on Pop Os 20.04.

flolivaud:
ALT+F2 then type r, and press enter is a simple workaround while we are waiting a fix about this issue.

PR #271, opened 2020-05-04
Fixed interaction with custom shell themes

### LINKED ISSUE #230: Pop-Shell disrupts theme use

NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
PRETTY_NAME="Manjaro Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.manjaro.org/"
SUPPORT_URL="https://www.manjaro.org/"
BUG_REPORT_URL="https://bugs.manjaro.org/"
LOGO=manjarolinux

 
 
commit:
16b4c33a8542969f19471b13b90764575a45ccf9

**Issue/Bug Description:**
it appears that while using a non-default theme, this disrupts the look of my current theme to be a mix of the default theme (Adwaita) and my theme of choice, Orchis (https://github.com/vinceliuice/Orchis-theme)

Launch with non default theme, can be fixed if theme is disabled then re-enabled or if the system is locked and then logged back into.

**Expected behavior:**
No effect on the theme.

### LINKED ISSUE #231: Pop Shell does not respect shell theme choice
 NAME=Fedora
ID=fedora
VERSION_ID=32
VERSION_CODENAME=""
PLATFORM_ID="platform:f32"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
VARIANT="Workstation Edition"
VARIANT_ID=workstation

 
 
[commit](https://github.com/pop-os/shell/commit/16b4c33a8542969f19471b13b90764575a45ccf9)

**Issue/Bug Description:**
when using non default shell theme, pop shell reverts my choice of theme to system default shell theme on session start(restart or login)

Select a non default shell theme and logout or restart, pop shell will revert the theme to default

**Expected behavior:**
Shell theme should be as it was before login/restart

jambonmcyeah:
Combines my previous PR #269 with @scaryrawr 's PR #270

The reason for this is addressed in the comments on the comments in #270

Closes #231, Closes #230

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -34,6 +34,8 @@ const { layoutManager, loadTheme, overview, panel, setThemeStylesheet, screenShi
 const Tags = Me.imports.tags;
 
 enum Style { Light, Dark }
+const STYLESHEET_PATHS = ['light', 'dark'].map(get_stylesheet_path);
+const STYLESHEETS = STYLESHEET_PATHS.map(Gio.File.new_for_path);
 
 interface Display {
     area: Rectangle;
@@ -166,7 +168,7 @@ export class Ext extends Ecs.System<ExtEvent> {
 
         this.load_settings();
 
-        this.register_fn(() => this.load_theme(this.current_style));
+        this.register_fn(() => load_theme(this.current_style));
 
         this.settings.int.connect('changed::gtk-theme', () => {
             this.register(Events.global(GlobalEvent.GtkThemeChanged));
@@ -391,7 +393,6 @@ export class Ext extends Ecs.System<ExtEvent> {
         }
     }
 
-    load_theme(style: Style) { load_theme(style === Style.Dark ? 'dark' : 'light') }
 
     monitor_work_area(monitor: number): Rectangle {
         const meta = global.display.get_workspace_manager()
@@ -677,7 +678,7 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     on_gtk_theme_change() {
-        this.load_theme(this.settings.is_dark() ? Style.Dark : Style.Light);
+        load_theme(this.settings.is_dark() ? Style.Dark : Style.Light);
     }
 
     on_show_window_titles() {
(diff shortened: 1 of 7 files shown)


src/search.ts
```diff
@@ -29,7 +29,7 @@ export class Search {
     ) {
         this.select_cb = select;
         this.dialog = new ModalDialog({
-            styleClass: "pop-shell-search",
+            styleClass: "pop-shell-search modal-dialog",
```
isantop, line 32: I like that we're using existing classes where they make sense here.

src/search.ts
```diff
@@ -170,13 +182,13 @@ export class Search {
 
                 let label = new St.Label({
                     text: title,
-                    styleClass: "pop-shell-search-label",
+                    styleClass: "pop-shell-search-label list-search-result-title",
```
isantop, line 185: Here (and the next one) also show good use of existing, supported style classes.

src/extension.ts
```diff
@@ -1347,13 +1347,30 @@ function find_unused_workspace(): [number, any] {
...
-        let theme = Lib.dbg(Me.path + "/" + stylesheet + ".css");
-        setThemeStylesheet(theme);
-        loadTheme();
-        return theme;
+        const themeContext = St.ThemeContext.get_for_stage(global.stage);
+
+        // Get the existing shell theme if exists
+        const existingTheme = themeContext.get_theme();
+
+        // Load the pop theme style sheet
```
isantop, line 1358: I would prefer for this comment to read "Load the Pop Shell style sheet", as it's not really the pop theme we're loading here. This is a nitpick and not something I'd say needs to get done before merging, but it would be nice to have.

isantop:
Pushed a copy of this branch into our repo for building/CI purposes. Please merge this PR instead of that one.

src/search.ts
```diff
@@ -170,13 +182,13 @@ export class Search {
...
                     y_align: Clutter.ActorAlign.CENTER
                 });
 
                 label.clutter_text.set_ellipsize(Pango.EllipsizeMode.END);
 
-                let container = new widgets.Box({ styleClass: "pop-shell-search-element" })
+                let container = new widgets.Box({ styleClass: "pop-shell-search-element button" })
```
isantop, line 191: This probably should not be a `button` class, as it makes the list of items in the Shell launcher look very strange.

jambonmcyeah: Fixed

src/extension.ts
jambonmcyeah: Fixed

jambonmcyeah:
Completed all the requests

src/search.ts
```diff
@@ -170,13 +170,13 @@ export class Search {
...
                     text: title,
-                    styleClass: "pop-shell-search-label",
+                    styleClass: "pop-shell-search-label list-search-result-title",
                     y_align: Clutter.ActorAlign.CENTER
                 });
 
                 label.clutter_text.set_ellipsize(Pango.EllipsizeMode.END);
 
-                let container = new widgets.Box({ styleClass: "pop-shell-search-element" })
+                let container = new widgets.Box({ styleClass: "pop-shell-search-element list-search-result" })
```
isantop, line 179: This looks better in Pop and Adwaita, but in many themes it causes the text to become unreadable. I think the existing styling present for these widgets was fairly theme-agnostic and resulted in guaranteed-readable text.

jambonmcyeah:
I addressed all your requests. While fixing the issues I noticed a bug with stylesheets only loading for the first time and fixed it. Also added detection for shell theme changes instead of relying on gtk theme changes, since they can be changed separately.

jambonmcyeah:
There's an issue with using custom styling, which is themes that style the shell dark even on light variants such as Adwaita and Plata don't work well. I removed the `color` in the css to make them work better, however it still doesn't work great. This is why I originally set the search elements to `button` to get around this issue.

jambonmcyeah:
The real solution to this problem is to implement search items using `St.Button` which is the [implementation by gnome-shell](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/master/js/ui/search.js#L37), instead of detecting arrow keys manually and faking focus. A proper implementation would have mouse support and would be compatible with themes. Yet I don't have nearly enough expertise on gtk and gnome to implement this properly, so I guess we have to live with this hack for now.

isantop:
@mmstick Some notes about the implementation of the Launcher we might consider :point_up:

isantop:
@jambonmcyeah Can you post a screenshot of the issues you are seeing with the stock styling? I worked pretty hard to try and get the styling to be fairly theme-neutral while also looking good.

jambonmcyeah:
Should be fixed now, I tested all the themes you mentioned and they all work fine

isantop:
@pop-os/quality-assurance Testing branch for this is now `jambonmcyeah-shell-theme-fix`

nicolae-stroncea:
This also fixes #227 and #216

Issue #155, opened 2020-04-14
Floating windows in tiling mode are not resizeable in window management mode


leviport:
If I'm in tiling mode, then pull a window out of the tile grid with Super+G to make it float, then enter window management mode with Super+Enter, Shift + arrow keys do not resize this floating window. I would expect the window to be resizeable like it is when "Tile windows" is turned off.

PR #282, opened 2020-05-04
fix: Autotile keyboard resize behavior is now consistent

mmstick:
Left / Right and Up / Down will now always decrease / increase. Position of a window will no longer dictate how it will resize.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1058,7 +1058,6 @@ export class Ext extends Ecs.System<ExtEvent> {
 
     size_signals_block(win: Window.ShellWindow) {
         this.size_signals.with(win.entity, (signals) => {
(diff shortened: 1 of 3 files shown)


mmstick:
Found an issue with this. Fixing it

Serkan80:
When I float an app in autotile mode, I cannot move it with my arrows keys. 
When I do that, it swaps with apps. 
And when I try to resize it, the app goes back into non floating mode and then the resizing starts the apply.

leviport:
@Serkan80 Thanks for the report. We're tracking that here: https://github.com/pop-os/shell/issues/155

mmstick:
Closing this because the fix I'm doing for reflows will also be fixing this.

PR #283, opened 2020-05-04
Update TESTING.md

WatchMkr:


TESTING.md
```diff
@@ -14,33 +14,81 @@ Note that because the logs are from GNOME Shell, there will be messages from all
...
+
+## With tiling enabled
+
+### Tiling
+
+- [ ] Super direction keys changes focus in the correct direction
+- [ ] Windows moved with the keyboard tile into place
+- [ ] Windows moved with the mouse tile into place
+- [ ] Windows swap with the keyboard (test with different size windows)
+- [ ] Windows can be resized with the keyboard (Test resizing four windows above, below, right, and left to ensure shortcut consistency)
```
leviport, line 27: We may want to add something about trying to resize windows that are in different positions. I assume a grid of 4 and trying each quadrant would be sufficient.

TESTING.md
```diff
@@ -14,33 +14,81 @@ Note that because the logs are from GNOME Shell, there will be messages from all
...
+- [ ] Windows can be resized with the mouse
+- [ ] Minimizing a window detaches it from the tree and re-tiles remaining windows
+- [ ] Unminimizing a window re-tiles it into its previous location
+- [ ] Maximizing removes the active hint and covers tiled windows
+- [ ] Umaximizing adds active hint and re-tiles into place
```
leviport, line 32: This currently doesn't always re-tile the window into the space it was formerly in. Should we add `(Not required, currently has a bug)`?

TESTING.md
WatchMkr: That's already there
- [ ] Windows can be resized with the keyboard (Test resizing four windows above, below, right, and left to ensure shortcut consistency)

TESTING.md
```diff
@@ -14,33 +14,81 @@ Note that because the logs are from GNOME Shell, there will be messages from all
...
 - [ ] The overlay hint correctly highlights the selected window
+- [ ] t: executes a command in a terminal
+- [ ] : executes a command in sh
+- [ ] = calculates an equation
+
+### Window Titles
+
+- [ ] Disabling window titles in Firefox works (Check debian and flatpak packages)
+
+## With Tiling Disabled
```
brs17, line 80: Perhaps adding a check that half tiling still works with tiling disabled would be a good test as well.

TESTING.md
leviport: Ah, misinterpreted the directions as "resize the windows in these directions". Sounds good then.

TESTING.md
WatchMkr: This is a living document that will constantly change. Approve the PR and add test cases as deemed necessary.

TESTING.md
```diff
@@ -14,33 +14,81 @@ Note that because the logs are from GNOME Shell, there will be messages from all
...
-
-- [ ] Windows can be rearranged with the mouse and keyboard
-- [ ] Windows can be swapped with the mouse and keyboard
-- [ ] Try all forms of resizing a fork with the mouse and keyboard
-- [ ] Minimizing a window should detach it from the tree
-- [ ] Fullscreening, unfullscreening, maximizing, and unmaximizing are all working
-- [ ] Restarting the extension; or toggling auto-tile; correctly handles minimized, maximized, fullscreen, floating, and non-floating windows
+- [ ] Windows move across displays in adjustment mode with direction keys
+- [ ] Windows move across displays with the mouse
+- [ ] Changing the primary display moves the top bar. Window heights adjust on all monitors for the new position.
```
leviport, line 59: Are we sure this doesn't currently have a bug? It doesn't seem to be working correctly for me. I'll file an issue about it.

Issue #198, opened 2020-04-18
Extending Launcher?


tinycrimsonpandas:
Hey there!

I've been working with the shell all day, and I really think this project is excellent. Thank you very much for all of the hard work.

One of my favorite parts is the launcher, it's a really intuitive way to navigate between the different tiles and launch applications. I also spend a lot of time on Mac OS where I've used quicksilver and, most recently, Alfred. 

There are things that I wish it could do, like open a URL in the default browser. I don't know your design concept around the scope of the launcher, but it might be interesting to explore extensions to enable different actions in the launcher.

AdamoPrincipe:
I use the Ulauncher for that.
https://ulauncher.io/

@tinycrimsonpandas maybe you can try that.

dhollinger:
@AdamoPrincipe Ulauncher is great (I use it, too) but i'd rather not have two launchers running. So, either the ability to extend the Pop Shell Launcher or disable it entirely (in favor of something like uLauncher) would be great.

mmstick:
Discussions on possible implementations are welcome

tinycrimsonpandas:
I agree that I wouldn't really want to operate two different launchers.

Perhaps there are three options:

- Extend the pop_shell launcher to allow extensions
- Provide a few additional options in the Pop shell launcher (URLs, perhaps other depending on user testing)
- Create an extension for uLauncher that can controll pop_shell? Or even perhaps then even include uLauncher as part of pop_shell?

If I wanted to take a look at the launcher code as it exists today, where in the source can I find that?

not7cd:
This file should interest you
https://github.com/pop-os/shell/blob/1d9c5b8d7e34763ce2f7b172ee879256e652488d/src/launcher.ts

I would like to create a custom mode, which would invoke a custom command.
Custom mode `g:$1` would run the custom command like `firefox -search "$1"`

nicolae-stroncea:
@not7cd, @tinycrimsonpandas you can change the following things in src/launcher.ts:

### Make a google search by typing *gg* in pop shell launcher

1. Change the MODES variable to the following: `const MODES = [':', 't:', '=','gg'];`
2. Inside of apply method in launcher.ts, add this as the last if check:
```
 else if (id.startsWith('gg')) {
    const expr = id.slice(2).trim().toString();
    const query = 'https://www.google.com/search?q=' + expr.replace(/ /g, '+');
    const cmd = 'xdg-open ' + query;
    log.debug(cmd);
[...]

PR #296, opened 2020-05-06
Add mode extensions to launcher

scaryrawr:
- Add web search mode triggered by '?'
- Add launcher search engine settings entry to schema
- Add launcher search engine settings to extension perf pane
- Add type interface for extending launcher
- Allow live preview for launcher extensions
- Change evaluate to a live preview mode

Targets #198

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -64,16 +64,6 @@
             <summary>Focus right window</summary>
         </key>
 
-        <key type="as" name="focus-monitor-left">
-            <default><![CDATA[['<Primary><Super>Left','<Primary><Super>h']]]></default>
-            <summary>Focus right monitor</summary>
-        </key>
-
-        <key type="as" name="focus-monitor-right">
-            <default><![CDATA[['<Primary><Super>Right','<Primary><Super>l']]]></default>
-            <summary>Focus right monitor</summary>
-        </key>
-
         <!-- Launcher -->
         <key type="as" name="search">
             <default><![CDATA[['<Super>slash']]]></default>
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -173,5 +173,11 @@
             <default><![CDATA[['<Primary>Right','<Primary>l']]]></default>
             <summary>Swap window right</summary>
         </key>
+
+        <!-- Launcher Options -->
+        <key type="s" name="search-engine">
+            <default>"https://www.duckduckgo.com/?q="</default>
+            <summary>Search engine query string</summary>
+        </key>
     </schema>
 </schemalist>
--- a/src/launcherext.ts
+++ b/src/launcherext.ts
@@ -71,17 +71,19 @@ export class CalcLauncher implements LauncherExtension {
     }
 
     search_results(expr: string): Array<[string, St.Widget, St.Widget]> | null {
+        const icon_size = this.search?.icon_size() ?? DEFAULT_ICON_SIZE;
+
         const item: [string, St.Widget, St.Widget] =
             [
                 `=${evaluate(expr).toString()}`,
                 new St.Icon({
                     icon_name: 'x-office-spreadsheet', // looks like calculations?
-                    icon_size: this.search?.icon_size() ?? DEFAULT_ICON_SIZE / 2,
+                    icon_size: icon_size / 2,
                     style_class: "pop-shell-search-cat"
                 }),
                 new St.Icon({
                     icon_name: 'accessories-calculator',
-                    icon_size: this.search?.icon_size() ?? DEFAULT_ICON_SIZE
+                    icon_size: icon_size
                 })
             ];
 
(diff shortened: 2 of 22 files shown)


schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
```diff
@@ -173,5 +173,11 @@
             <default><![CDATA[['<Primary>Right','<Primary>l']]]></default>
             <summary>Swap window right</summary>
         </key>
+
+        <!-- Launcher Options -->
+        <key type="s" name="search-engine">
+            <default>"https://www.google.com/search?q="</default>
```
scaryrawr, line 179: Would something like duckduckgo be better for a default?

src/launcherext.ts
```diff
@@ -0,0 +1,169 @@
...
+    /**
+     * Initializes the launcher extension
+     * @param ext Extension settings
+     * @param search Launcher instance
+     */
+    init(ext: Ext, search: Search): LauncherExtension;
+
+    /**
+     * Perform apply action
+     * @param text The currently typed text in input
```
scaryrawr, line 34: currently typed text minus the prefix.

src/launcherext.ts
```diff
@@ -0,0 +1,169 @@
...
+
+        this.search?.set_text(`=${value}`);
+
+        return true;
+    }
+
+    search_results = (expr: string): Array<SearchResult> | null => {
+        const item: SearchResult =
+        {
+            display: `=${evaluate(expr).toString()}`,
```
scaryrawr, line 76: Doing this as the user types actually generates a lot of logs...

src/launcherext.ts
```diff
@@ -0,0 +1,169 @@
...
+        return this;
+    }
+
+    apply(cmd: string): boolean {
+        let terminal = TERMINAL.get_or_init(() => {
+            let path: string | null = GLib.find_program_in_path('x-terminal-emulator');
+            return path ?? 'gnome-terminal';
+        });
+
+        spawnCommandLine(`${terminal} -e sh -c '${cmd}; echo "Press to exit"; read t'`);
```
scaryrawr, line 120: this is copied from the original mode switch, but `-c` is not supported by gnome-terminal, so this is broken :crying_cat_face:

src/launcherext.ts
scaryrawr: image

src/launcherext.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    apply(webSearch: string): boolean {
+        // xdg-open should use user's default browser
+        const cmd = `xdg-open "${this.get_query(webSearch)}"`;
+        spawnCommandLine(cmd);
+
+        return false;
+    }
+
+    search_results(webSearch: string): Array<SearchResult> | null {
```
scaryrawr, line 152: Not sure what the icons should be here.
image

schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
mmstick: It would indeed

mmstick:
Can you seperate each specific change into a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/)? Essentially, there should be one commit for each of your bullet points.

src/launcherext.ts
mmstick: Does it work if changing the `-e` to `-x`?

mmstick: Nevermind, this breaks Alacritty and other terminals

mmstick:
I don't think `?` would be ideal for web queries. I could see that character theoretically being used for searching help documentation. Perhaps `w` instead?

src/launcherext.ts
mmstick: I wonder if there would be value in display the search results of a query here, instead of the URL of the query.

scaryrawr: I think results would be an awesome enhancement. Being able to select the item and launch the browser to the page instead of just launching the query. I think it's a bit more complicated to figure out though since it starts adding web calls from launcher, and having a different design for web results.

scaryrawr: I was looking into having a web view be displayed, but it looks like it's not supported by Clutter. DuckDuckGo has an [instant answers api](https://duckduckgo.com/api), but it's not the same as full search results.

mmstick: It's not, but you could scrape the HTML output for titles & links and list search results that way

scaryrawr: I started looking into this, I can get the HTML using Soup, but there doesn't seem to be a decent XML Parser in GJS? It looks like there's [Rest.XmlParser](https://gjs-docs.gnome.org/rest07~0.7_api/rest.xmlparser#method-parse_from_data), but playing around in GJS, I can only see the root HTML node lol. I think we'd also have to add support for async search results in search since doing synchronous sessions would pause user typing I think.

mmstick: If it's a lot of work, it'd be best to wait for GJS to have support. I'm impatiently waiting for the day we can use Rust WebASM modules in GJS.

mmstick:
I will soon cherry pick commits from this PR which are ready for inclusion

mmstick:
All but the web search was added. Closing as the original intent of this PR is now in the dev branch for review

PR #299, opened 2020-05-06
fix: Ignore decoration of Firefox windows via WM_CLASS value

mmstick:
This should be more reliable than watching the name of a window


mmstick:
Changes now in `dev_focal` branch

PR #323, opened 2020-05-08
Fix link mistake

leviport:
I put one link in the wrong spot and forgot the other link. This fixes it.


leviport:
Changes made irrelevant by https://github.com/pop-os/shell/pull/342

PR #342, opened 2020-05-14
Update TESTING.md

WatchMkr:
Add tests for new gnome settings search and duplicates in the launcher. Remove bugs listings that were enhancements or fixed. Removed notes and conditions for non-regressions which now work.


TESTING.md
```diff
@@ -84,7 +80,6 @@ Tasks for a tester to verify when approving a patch. Use complex window layouts
 ### Window Titles
 
 - [ ] Disabling window titles in Firefox works (Check debian and flatpak packages)
-  - This may make firefox windows move around randomly when toggled, and you may have to restart firefox for it to work properly, but it should work besides that.
```
brs17, line 87: This is still the case, and may be a good thing to leave in here so that testers immediately know this is not a regression.

Issue #259, opened 2020-05-03
Tray icon remains after disabling extension


daveriedstra:
**Distribution (run `cat /etc/os-release`):**
 ```
NAME="Pop!_OS"
VERSION="20.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.04 LTS"
VERSION_ID="20.04"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
LOGO=distributor-logo-pop-os
```
 
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):**

```
pop-shell:
  Installed: 0.1.0~1588274983~20.04~586efcc
  Candidate: 0.1.0~1588274983~20.04~586efcc
  Version table:
 *** 0.1.0~1588274983~20.04~586efcc 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main i386 Packages
        100 /var/lib/dpkg/status
```

**Issue/Bug Description:**

Tray icon still present after pop shell extension is disabled

**Steps to reproduce (if you know):**

1. Disable extension

**Expected behavior:**

Tray icon shouldn't be present when extension is disabled

![image](https://user-images.githubusercontent.com/3504098/80918050-f2d1be00-8d30-11ea-9000-6fbf2cbd7371.png)

nicolae-stroncea:
Can confirm same behaviour on Fedora 32

daveriedstra:
After a restart the icon is gone. This is similar to [my experience](https://www.reddit.com/r/pop_os/comments/gcjk5i/some_theme_settings_reset_on_login/) with #230 and #231 so might be related.

isantop:
I can also confirm this behavior on Pop_OS 20.04

rthery:
I believe I've located the source of the issue, the indicator is not destroyed in `disable()` (as suggested to do so in https://wiki.gnome.org/Projects/GnomeShell/Extensions/Writing).

I'll prepare a PR for it after testing today if all goes well!

PR #343, opened 2020-05-14
fix: Destroy panel indicator disabling extension

rthery:
Tested, disabling the extension properly removes the button indicator from the panel now.

Closes pop-os/shell#259

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1355,6 +1355,11 @@ function disable() {
         ext.keybindings.disable(ext.keybindings.global)
             .disable(ext.keybindings.window_focus)
     }
+
+    if (indicator) {
+        indicator.destroy();
+        indicator = null;
+    }
 }
 
 function find_unused_workspace(): [number, any] {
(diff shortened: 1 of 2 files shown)


mmstick:
Does it correctly re-enable on unlocking the screen?

rthery:
@mmstick Yes, just tested, the indicator is still here and functional after locking and unlocking.

Although I noticed that we don't clear the Active Hint (if the option is enabled when we disable the extension). I checked the issues, none is about that specific one yet.
How would you prefer to handle this ? I can create a new issue and look into it later for another PR.

mmstick:
That's good. GNOME Shell calls the `disable()` method on locking the screen, so I was worried that it would break it on locking / unlocking. Anything that can be disabled should be disabled, so long as we get it working again as it was after unlocking.

rthery:
Alright, I've fixed the issue with the active hint as well then!
Visually, toggling the extension doesn't leave anything behind now from what I can test.

mmstick:
Closing as it's now in the dev branch for review

mmstick:
Thanks for the PR!

rthery:
@mmstick My pleasure, it was my first contribution to a Linux related open source project, and I'm glad it was for Pop!_OS. Continue your awesome work at System76 and I hope I'll be able to contribute more in the future :)

PR #357, opened 2020-05-16
toggle_tiling keyboard shortcut

timrichardson:
Added a new keyboard short cut 'toggle-tiling'.
to do this, added method toggle_tiling to Extension. 
This new method is a relocation of the toggle code from panel_settings.
I have used Super-Y as the default shortcut because it was the first key I thought of that didn't seem to be mapped already. 

Warning. This is the first time I have used typescript, and it's the most Javascript I've done in five years.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -313,6 +313,53 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     // Extension methods
+    
+    toggle_tiling() : any {
+
+        if (this.auto_tiler) {
+            Log.info(`tile by default disabled!`);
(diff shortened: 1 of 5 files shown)


src/extension.ts
```diff
@@ -313,53 +313,6 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     // Extension methods
```
timrichardson: realised that the methods are alphabetically sorted

timrichardson:
I will resubmit this tidied up and rebased to master_focal

PR #359, opened 2020-05-16
Issue 321 (rebased) Add keyboard shortcut for toggling auto-tiling

timrichardson:
moved toggle code from panel_settings and into extension (new method toggle_tiling), so it can be more generally used. Added new keybinding. I chose Super-Y because it is not used at the moment. I did not think or care much about the default assignment. 
This is the first time I've used typescript, gnome extensions and it's the most javascript I've done in years. Thanks for making it so easy to get started.


mmstick:
Thanks for the PR!

timrichardson:
Note: it is not updating the state of the toggle, I will work out how to do that.

Issue #47, opened 2020-03-04
Old Keyboard Shortcuts are Lost W/out warning


Gryzor:
### Context
I wanted to try this to start offering early feedback and to start getting used to what I think is going to be amazing so I asked around and @mmstick got me hooked up right away (thanks!). 

I'm on Pop 19.10 and had to do `npm install -g typescript` (I already had `npm` set up). 
`tsc --version` now returns `Version 3.8.3`

This is an Oryx Pro 2019.

### Problem
After running the `rebuild.sh` script, I noticed all my mostly default familiar Pop!_OS shortcuts to move windows around no longer worked. This is obviously expected (that there will be new _conflicting_ shortcuts with a regular Gnome Session), but what was not necessarily expected was that I was going to "lose" my original shortcuts. 

Perhaps there's a way to mitigate this (warn users? allow to "save" them?)

### Expected
That there was a way to restore my shortcuts the way they were before.

### Why am I filing this?
At the request of "shpurk" on mattermost who said: (And I quote):

> Maybe you should file an issue about this, the fact that you didn't think to do this/etc, when people upgrade to 20.04, there will probably be a lot more people in the same spot you are in.

Thanks!

mmstick:
This project won't be responsible for altering keyboard shortcuts. The rebuild script is for development / testing purposes only.

mmstick:
I believe we'll have a warning in the OS upgrade's changelogs about the changes we're making.

brs17:
In the changelog when updating to Pop!_OS 20.04, we mention the changing of keyboard shortcuts.

brs17:
Thank you for posting the issue @Gryzor

nicolae-stroncea:
I think this is still an issue for beta testers, and people running it on other distributions for their first time. They don't have the Pop 20.04 Changelog, so they don't expect their shortcuts to be modified.

It would probably be good to warn them in the README, and even better by prompting for a confirmation in rebuild.sh: "Pop Shell will change your shortcuts. Are you sure?". Type y/n.

Ideally the confirmation would be saved somehow so you don't get prompted again, but the simplest implementation would probably be. rebuilding with: `./rebuild.sh -y`.

PR #375, opened 2020-05-19
Shortcut override confirmation in rebuild.sh

nicolae-stroncea:
Ensures that people who are building this for other distros are aware that it will override their shortcuts.

Rebuild.sh now creates a hidden file when user confirms overriding shortcuts, and rebuild script checks for the hidden file every time its run.

Targets  #47

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ debian/*
 !debian/control
 !debian/copyright
 !debian/rules
+.confirm_shortcut_change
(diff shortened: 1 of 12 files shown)


olejorgenb:
Thanks. Very jarring that a ton of shortcuts are simply removed with no undo option. I know it's documented in the README, but many people are sloppy readers when it comes to instructions  :)

Even better would be to do a `dconf dump` of the altered keys so it's possible to restore the shortcuts.

mmstick:
Merged into dev_focal. Thanks!

PR #395, opened 2020-05-27
fix: Add check for wayland before invoking xprop

### LINKED ISSUE #166: Pop Shell makes Gnome unresponsive after login [Appears to be Wayland only]
As the title says, after login no command affects gnome and the only way to restart is to hard reset the machine into a text only interface and deleting the extension from there.

Also I noted that my clock did not change whilst gnome was unresponsive

Manjaro Gnome 3.36
Wayland

scaryrawr:
- Add `utils.is_wayland()` check in `set_hint`
  - Without check, wayland appears to freeze when a window is created
- Tiling appears to work with the early return

Fixes #166

### DIFF
--- a/src/xprop.ts
+++ b/src/xprop.ts
@@ -3,6 +3,8 @@ const Me = imports.misc.extensionUtils.getCurrentExtension();
 
 import * as lib from 'lib';
 
+import * as utils from 'utils';
+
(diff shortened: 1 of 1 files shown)


scaryrawr:
oops, apparently didn't switch to wayland, issue still persists.

PR #396, opened 2020-05-27
fix: do not add active_hint for windows with skip_taskbar

folke:
Windows with state `skip_bar` should not get the `active_hint`.

This resolves issues with launchers like ulauncher, or terminal dropdowns like guake.


mmstick:
Thanks!

PR #409, opened 2020-06-09
Blacklist Kotatogram's mediaviewer window from being tilable

RememberTheAir:
This pull request will blacklist Kotatogram Desktop's image viewer window from being tilable.

Kotatogram is a fork of Telegram Desktop, which already appears in the blacklist


mmstick:
Thanks!

PR #419, opened 2020-06-14
Fix moving window between monitors hotkey.

AlphaCluster:
Need to disable the default move to monitor up and down so that moving workspace up and down actually works.


mmstick:
Thanks!

PR #420, opened 2020-06-14
Add JetBrains Toolbox to blacklist

AlphaCluster:
JetBrains Toolbox doesn't tile well and the window acts more like a dropdown for the taskbar.


mmstick:
Thanks!

Issue #411, opened 2020-06-09
Empty space left behind after minimizing for some apps


lucaspallez:
**Distribution (run `cat /etc/os-release`):**
 Pop!_OS 20.04
 
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):**
From apt:
pop-shell: 0.1.0\~1589925465\~20.04~ff5f374

**Issue/Bug Description:**
In tiliing mode, empty space is left behind when minimizing some windows every time when using minimize button in top bar (it works properly when using for example dash to panel to minimize).
This happens with these apps and probably more that I don't know of:

-Spotify (deb)
-Programs using wine (deb)
-Discord (deb)
-Slack (deb)
-Lotion (Notion wrapper: https://github.com/puneetsl/lotion) (from source)
-Vnc Viewer (deb)
-VirtualBox (deb)
-Zoom (flatpak)

Maximizing the windows works, but when back to normal size the problem persists.
Turning tiling mode off and on again solves the problem.
Closing the app instead of minimizing it also removes the empty space.

**Steps to reproduce (if you know):**
Open one of these apps and minimize it, it should work the same as on my machine. Not modifications were made on my part to pop-shell.

matguidi:
Another app that has this problem is Enpass.

jmmaranan:
Duplicate of #309

bvhme:
I see this in Mattermost too. I've made in issue on their list as well: https://github.com/mattermost/desktop/issues/1325

jmmaranan:
@bvhme - the Mattermost minimize issue you described is also fixed in the dev_focal branch. Luckily, the reference PR or issue that one of the commenters described in mattermost/desktop#1325 would probably fix the close (x) button issue I am seeing in Mattermost display.

jmmaranan:
@lucaspallez - can you pull the latest apt-release and verify if it is fixed on your end?

lucaspallez:
It is indeed fixed! @jmmaranan
Thank you.

jmmaranan:
@lucaspallez - do you mind closing this issue?

matguidi:
> @lucaspallez - can you pull the latest apt-release and verify if it is fixed on your end?

Sorry, for those who use pop-shell in a non-Debian-based environment like me, is the release you refer to the same as the current dev_focal?

jmmaranan:
[...]

Issue #346, opened 2020-05-15
Two Tiled Chromium windows on one monitor.When one is  minimised, the other is not retiled, and then when the first is unmiminised, it goes on top of the other window


timrichardson:
**Distribution (run `cat /etc/os-release`):**
 NAME="Pop!_OS"
VERSION="20.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.04 LTS"
VERSION_ID="20.04"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
LOGO=distributor-logo-pop-os

 
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):**
tim@keynes:~$ apt policy pop-shell
pop-shell:
  Installed: 0.1.0~1589481653~20.04~87a5f44
  Candidate: 0.1.0~1589481653~20.04~87a5f44
  Version table:
 *** 0.1.0~1589481653~20.04~87a5f44 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main i386 Packages
        100 /var/lib/dpkg/status
tim@keynes:~$ 

**Issue/Bug Description:**
Two Tiled Chromium windows on one monitor, say Left and Right. When Right is  minimised, Left is not retiled, and then when the first is unmiminised, it goes on top of the Left.
It is as if Left is treated as not existing, when Right is minimised 

**Steps to reproduce (if you know):**
I have multiple Chromium profiles (ie "Accounts")
I start one profile and then from  that open a second profile. This creates a new Chromium windows, and tiling works. I have Left and Right.
When I minimise Right, what I describe in  the Description happens.
When I restore, the restored windows tiles as if there were no windows there already. I can not get it to acknowledge the existing window.
I can reproduced this every time.
I have three monitors. This is happening on monitor 2, a 1980x1200 landscape monitor. 

However,  at some point, tiling seems to reset. If I had a third window to the desktop (a terminal window for instance), it seems to correct. But this is hard to reproduce., the behavior does not feel deterministic. 

**Expected behavior:**
[...]

Issue #309, opened 2020-05-06
shell holds a blank area when minimizing the discord and terminator


cansozbir:
**Distribution (run `cat /etc/os-release`):**
NAME="Pop!_OS"
VERSION="20.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.04 LTS"
VERSION_ID="20.04"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
LOGO=distributor-logo-pop-os

 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):**

pop-shell:
  Installed: 0.1.0~1588624471~20.04~dfb378c
  Candidate: 0.1.0~1588624471~20.04~dfb378c
  Version table:
 *** 0.1.0~1588624471~20.04~dfb378c 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main i386 Packages
        100 /var/lib/dpkg/status

**Issue/Bug Description:**
When I minimize Discord app, it does not change the view, the shell is acting like the app is not minimized and holds it as a blank area. 

**Steps to reproduce (if you know):**
1. Open a window
2. Open the Discord app
3. Minimize the Discord app

**Expected behavior:**
It should expend the first window. 

**Other Notes:**
I installed discord from pop-shop (.deb version)
There is no problem when minimizing terminal (I used pop os's default terminal on the video to show expected behavior), but the same problem with Discord is happening with Terminator app too. 

![pop-shell-bug](https://user-images.githubusercontent.com/32648255/81233669-cbfcdb80-8fff-11ea-9903-1ec6e6fba848.gif)

TC68:
I am also experiencing this for other applications including Firefox and Thunderbird.

NAME="Pop!_OS"
VERSION="20.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.04 LTS"
VERSION_ID="20.04"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
[...]

Issue #237, opened 2020-04-30
Minimizing window does not retile layout


nicolae-stroncea:
**Distribution (run `cat /etc/os-release`):**
 
Fedora 32, Gnome 3.36, Xorg
 
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):**

57fd88ce38b1ec5e4f06db051f881dd9fd053943

**Issue/Bug Description:**

When you minimize a window, the layout isn't retiled and the space remains empty where the window used to be.

**Steps to reproduce (if you know):**

* Minimize a window

mmstick:
And fixed

jmmaranan:
@nicolae-stroncea - do you still remember what types of applications are not able to minimize? QT, Electron, or GTK based?

jmmaranan:
@mmstick - looking at the apps mentioned in #290 - this is fixed. Android studio, etc

nicolae-stroncea:
@jmmaranan  unfortunately I don't remember which apps that occured for. I did a couple of small tests right now for Electron and GTK and they seemed to work fine

jmmaranan:
@mmstick, I think this can be closed too

PR #421, opened 2020-06-14
fix(autotiler): handle minimize on_grab_end

jmmaranan:
Update 1: Also checked #346
***
Fixes for #309, #411 and not sure on #237.

Tested these apps (can replicate and then are now fixed):
- Terminator
- Discord
- MS Teams Preview
- Spotify
- Slack

Electron apps or some apps that manage their own window title bars do not
send minimize events, but rather PopOS shell treats them as a grab event.

When `on_grab_end` event is  sent by these apps and the meta.minimized
attribute is set, re-using the default `extension->on_minimize()` handler should fix the empty space being left by these types of apps. Which should properly cleanup and call the `autotiler -> detach_window()` function.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -675,6 +675,11 @@ export class Ext extends Ecs.System<ExtEvent> {
 
         this.size_signals_unblock(win);
 
+        if (win.meta && win.meta.minimized) {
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks!

PR #428, opened 2020-06-17
Fix closed applications dash

### LINKED ISSUE #220: Closed Applications Appear Running in the GNOME Dash and Activities View (Can't open app)
**How did you upgrade to 20.04? (Fresh install / Upgrade)**
Fresh install
 

 
Not sure about that.

**Issue/Bug Description:**

When I enable the "Tile Windows" feature and then switch workspaces while closing some apps (not evening running in the background), the dock (or the favorites menu) still shows them active.

As you can see in the image here: https://imgur.com/QPIQZbq

Even though Firefox and the file manager has been closed (not even in the background, I repeat). The icon still shows up in the menu as **active** for some reason.

This only happens when I have the tile windows feature enabled and try to switch workspaces.

1. Enabled **Tile Windows.**
2. Open any app in the first workspace
3. Open any app in the second workspace.
4. Close an app and make sure it's not running in the background.
5. Check your dock/menu.

**Expected behavior:**

It should not show that dot thing under the app icon (which indicates that the app is still active).

**Other Notes:**

Feel free to edit the **title** of this issue, I wasn't quite sure how to put it.

jmmaranan:
Was still reproducible on #405. This additional change fixes #220.

Test case 1 (happens only on Tiling Mode):
1. Click icon Files on Dash to Dock
2. Right-click New Window from Dock.
3. Focus on one of the Files window, CTRL+Q
4. Should not leave any window dots on Dock.

Test Case 2 (happens only in Tiling Mode):
1. Toggle Title Bar on Firefox Customize page.

This is probably a question to PopShell devs on what this call is for:
```
// window.meta.change_workspace_by_index(workspace, false);
```
If this is disabled, the closing of apps via CTRL+Q does not leave any ghost dots in Dash to Dock nor toggle of Firefox's Title Bar in Customize.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ debian/*
 !debian/copyright
 !debian/rules
 .confirm_shortcut_change
+.vscode
--- a/src/forest.ts
+++ b/src/forest.ts
@@ -75,7 +75,11 @@ export class Forest extends Ecs.World {
             const window = ext.windows.get(entity);
             if (!window) continue;
 
-            window.meta.change_workspace_by_index(workspace, false);
+            
+            if (workspace) { 
+                // do nothing for now
+                // window.meta.change_workspace_by_index(workspace, false);
+            }
 
             move_window(ext, window, r.rect, () => { });
         }


rengare:
This means that the window is closed but when window.meta.change_workspace_by_index is called, pop-shell holds a reference to a window which should be killed. There should be a method to check if the kill signal was send

jmmaranan:
@rengare - I guess the question is that if this method call is really necessary. Because it looks like the method itself is from gnome-shell js api. I don't see any other regressions when this is not done.

rengare:
@jmmaranan that's true, I didn't notice anything strange either

mmstick:
Could you rebase this into a single conventional commit? I'll merge when ready. Thanks!

jmmaranan:
@mmstick - done

PR #429, opened 2020-06-17
feat: Stacked tiles mode

mmstick:


jackpot51:
Any reason not to have a binding for stacking without having to enter tiling mode? I added one called toggle-stacking-global and added it to keybindings.ts and it seems to work ok

mmstick:
@jackpot51 No reason — just need a good shortcut for it.

jacobgkau:
I know the primary input method for navigating Pop Shell is meant to be the keyboard, but I noticed pretty quickly after activating stacking that the tabs are a bit difficult to click. It seems like only the top half of the tab is clickable, since the bottom half is close enough to the top of the window that it tries to resize instead.

stacking-1

stacking-2

jackpot51:
Good find @jacobgkau.

@mmstick I don't see any difference between Super and Super+S. I wish we could re-purpose that

ids1024:
> I don't see any difference between Super and Super+S. I wish we could re-purpose that

If we don't anticipate many people complaining, that could be disabled by default [in pop-session](https://github.com/pop-os/session/blob/master/debian/pop-session.gsettings-override). It does seem kind of redundant; not sure if many people use it.

mmstick:
I think the best solution for the grab handle overlapping with the tabs in the stack's UI actor is to have the tabs be a separate UI actor which is always kept above the active window. I will try this out.

I've pushed a commit which adds the `<Super>S` shortcut, and the rebuild script will unset `toggle-overview`.

jackpot51:
@mmstick can you prepare a pop-session change, under a branch stacking_focal, that unsets toggle-overview?

WatchMkr:
@jacobgkau post light mode screenshots here as well. I'd like to have Kate take a look at our color choices for tabs.

mmstick:
There's currently no different in light or dark themes

WatchMkr:
@mmstick it's to get an idea of how it looks in light mode and determine if we need different colors or if we can choose colors that will work well in both.

jackpot51:
It would be nice if the tab had the window title instead of the app name. Hard to distinguish between a dozen different "Terminal" or "Firefox Web Browser" tabs

kazen117:
I certainly agree with Jacob that the tabs should be taller, for visibility reasons in addition to usability. As for the color, these seem to be pulling the same contrast, but that doesn't look like the lowlight/shade of our orange. It also looks like the highlighted tabs are both pulling the dark mode highlight of our orange. So these don't look like they're defaulting to the Pop theme colors and contrast. Is that correct? Where are these colors coming from?

kazen117:
For context, in case it may be missing for anyone here, we have tints of our brand grey color; we've articulated tints and shades for all our major brand colors and additional colors necessary for various other UI needs. (We also now have names for our colors, but I won't add an extra layer for you all to remember just yet.) 

image

colors 1

I think one of these lighter shades may be of use here for deselected tabs. Maria and I are working on mocking this up right now.

kazen117:
@isantop @WatchMkr given that the orange highlight contrast is an important and common shift between light and dark mode, we can keep the deselected tab color the same, but the highlight should really change to remain consistent with each theme. This is the proposed color theory, using #9b8e8a for deselected tabs and #fbb86c for selection in dark mode / #ffad00 for selection in light mode.

light-mode-tabs

dark-mode-tabs

WatchMkr:
Color looks great. Too chunky though. The top bar on GNOME apps is large to accommodate buttons and menus. These tabs should be similar to Firefox tab sizes which appear a bit taller than the current screenshots.

kazen117:
Sure. It was a bit exaggerated in response to how tight the original was—it would be great if we could emulate Firefox or Chrome-like tab design in the actual implementation! 

Take it purely for the color articulation and discard any context that isn't useful.

kazen117:
@WatchMkr actually, for the tabs, should we mock this up specifically, or is there a styling in the works already?

WatchMkr:
@mmstick would you like a mock-up from design? Up to you.

kazen117:
FWIW, Maria and I realized we should consider differentiation within a hierarchy, i.e. "if you have a bunch of Firefox tabs open and Firefox is a tabbed application among several, what does that look like?", so we did start on mockups already. But let me know if there's already a path forward for that.

WatchMkr:
Mockups welcome! Jeremy mentioned changing the description to the window title instead of the application name which will help differentiate between multiple firefox or terminal windows that are open. The tab would also then match the launcher description.

mmstick:
Yep. I got this working now, which also required tracking title property notifications for as long as the window is inside the stack. I haven't yet pushed the changes though.

leviport:
I'm seeing a bug relating to swapping a tile group with an untiled window. I start with an unstacked window and a stack of two windows, something like this: 
Screenshot from 2020-08-03 11-13-41
(I have two windows open above them to better show the issue, but it still happes without having the extra two windows open)

Starting with one of the stacked windows active, if I go into window management mode (Super + Enter), then swap the active window to the one beside it (Ctrl + right arrow, then Enter), the stack and lone window get swapped, but there is an issue with the tabs.
Screenshot from 2020-08-03 11-13-52

It looks like the space for the tabs remains where the tabs on the left were, and the tabs on the right now extend past the dimensions of where the lone window was.

jacobgkau:
With the most recent commit (009eb19), I'm still seeing the issue after swapping a stack with a window that Levi discovered.

This next one seems less critical, but I noticed that in a default quarter-tile (without resizing anything), just clicking (not dragging) on the top resize border of a stack can have undesirable consequences. On a 1080p display, the text becomes distorted (the tabs seem to move down a pixel or so, causing the distortion):

text-distortion

Video of reproducing this (it's a little difficult to see when not zoomed in): https://www.dropbox.com/s/8buarngpg7gw165/Screencast%20from%2008-03-2020%2004%3A01%3A59%20PM.webm?dl=0 

This does not seem to happen on a 1440p secondary display. On a 4k primary display, clicking (once again, not visibly dragging) sometimes causes the window to decrement in size: https://www.dropbox.com/s/x60azyzay1y3th5/decrement.webm?dl=0 

On one hand, I wouldn't expect either of these issues to affect keyboard-centric users. On the other hand, since the top window border's resize handle overlaps part of the tab, it's easy to click on the resize handle when you're trying to click between tabs. (I would personally prefer the resize handle to be at the top edge of the tabs, since it seems like it would be easier to avoid the top edge than to target the top edge, but that might just be me.)

mmstick:
Closing in favor of #481

Issue #447, opened 2020-07-02
When GNOME Control Center makes itself larger, other windows do not get smaller


jacobgkau:
**Distribution (run `cat /etc/os-release`):** Pop!_OS 20.04 LTS
 
 
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):** `0.1.0~1593273798~20.04~9e8ab29`

**Issue/Bug Description:** Some pages of the GNOME Control Center have larger minimum widths than others. If a page is selected and the window is too skinny to display it, GNOME Control Center horizontally grows to the selected page's minimum width. When tiling, other windows do not resize to compensate, and are overlapped by the GNOME Control Center window until G-C-C is resized again. (Other windows are also able to be grown in this state so they overlap even farther.)

**Steps to reproduce (if you know):** Open GNOME Control Center on the left side of the screen and another window on the right side of the screen. Select the WiFi settings and make G-C-C as skinny as possible, then select the Mouse & Touchpad settings.

**Expected behavior:** When G-C-C makes itself larger, other windows should shrink to compensate.

**Other Notes:**

jmmaranan:
It looks this is not fixed by the WM_HINTS change. I missed the step where I need to select each panel like Network, etc when the width is shrunk.

PR #448, opened 2020-07-03
feat(window): remove window hints via Gdk

jmmaranan:
There seems to be an API to adjust the window hints via GDK11. Looking for feedback @mmstick.

**Experiment**: change hints on tile, remove hints on detach. There has to be a decision on what is the smallest possible window to view, versus just splitting them all like i3?

**What works:**

- To users who want (the thrill) to resize the windows to as much as it can. Not sure who likes a 20 x 60 window.
- This makes it possible for pop-shell to control some boundaries on the size of some windows. #447 
- Edit Win Mode (`Super + Enter`), can now squeeze some windows.

**Issues:** 

- The shell responds slowly on closing windows.
- Dragging and dropping window edges using mouse is glitchy.
- During `Alt + F2, r` on windows that glitched will be closed by gnome-shell.

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -139,7 +139,8 @@ export class AutoTiler {
     detach_window(ext: Ext, win: Entity) {
         this.attached.take_with(win, (prev_fork: Entity) => {
             const reflow_fork = this.forest.detach(prev_fork, win);
-
+            const meta = ext.windows.get(win)
+            meta?.restore_window_hints()
             if (reflow_fork) {
                 const fork = reflow_fork[1];
                 this.tile(ext, fork, fork.area);
--- a/src/window.ts
+++ b/src/window.ts
@@ -15,7 +15,7 @@ import type { Ext } from './extension';
 import type { Rectangle } from './rectangle';
 
 // const GLib: GLib = imports.gi.GLib;
-const { Gdk, Meta, Shell, St } = imports.gi;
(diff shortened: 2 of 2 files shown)


mmstick:
Thanks! I'll look into these issues

mmstick:
Technically it should only restore hints when toggling floating, or disabling tiling. Detaching occurs when a window is removed from one place, and placed somewhere else in the tree.

jmmaranan:
@mmstick - nice! Thanks for the follow-up. The additional changes indeed fixed the closing issue.

jmmaranan:
@mmstick - the `restore hint` function does not seem to technically remove what the `override hint` did. From what I have read, the `Gtk.Window` instance is able to get the defaults by calling the `default_size` or `allocated_size`. Which should be done and stored prior to overriding the geometry hints. But a Gdk.Window is not a Gtk.Window nor vice-versa. I cannot find an API/function method in the library to traverse in between. 

This is what happens to the _tracked windows` after pop-shell _releases_ them (floating, or disabling tiling). On below, I tried to drag the right-hand side edge to shrink the window. Same with the bottom edges.

**Workaround** for user is to re-open the said affected windows. Not sure if that is going to be acceptable to some.

image
image

mmstick:
I actually already have a function in here that uses xprop to get the minimal sizes from the `WM_NORMAL_HINTS` key. We can get these values before setting the MIN_SIZE through Gdk

jmmaranan:
I'll let you know when I have tested it.

PR #481, opened 2020-08-05
Add stacked tiles feature to tiling mode

### LINKED ISSUE #150: Tabbed Mode
I love the Window Tiling feature. But I commonly find that I want to have a group of the same kinds of windows pinned to a tile area. Basically some sort of hybrid between `tiling` and `stacking`. 

Here is example:
- On the right hand side of the screen in a column I want a terminal session (full height of the screen).
- On the left hand side of the screen I want N number of code editors all pinned into the same column (not tiled but stacked). Ex: 4 instances of Visual Studio Code all stacked but "pinned" into the same tile area.  

I understand the goal with tiling is to make every window visible - but I often find myself using the feature more to create "spaces" where specific kinds of applications are held. 

Along these same lines - if I launch N+1 of the same code editor (say Visual Studio Code) I want it to automatically show up grouped with it's peers.

### LINKED ISSUE #488: Quake tile/Iconify option
1) Would it be possible to add a Quake style options [ Ex: Guake/Tilda ] for specific applications?
I've run into the situation where I'm constantly reshuffling my windows due to windows that I want to leave open due to frequent bursts of short usage.
I've been able to use terminals like tilda and tilix with Quake feature for drop down terminals.
I couldn't think of a more appropriate application to take on this option to use dropdown features for nautilus & other moderate usage apps.

1b) This would allow an app to not have to take up a tile

2) If quake does not fall into scope, can we have an iconify shortcut, where they can be shrunken to take up a postage stamp size
icon where it can be restored to size upon keyboard shortcut? 

2b)This stamp would show a keyboard cmd [Alt F12] to restore it. Either of these options would allow for keyboard control of applications.

2c) An option to shrink all other windows per monitor can also be available and would provide user option to restore the appropriate windows for current usage

### LINKED ISSUE #522: "move window one monitor to the right/left" doesn't work with tiling enabled

```
NAME=Fedora
ID=fedora
VERSION_ID=32
VERSION_CODENAME=""
PLATFORM_ID="platform:f32"
ANSI_COLOR="0;34"
LOGO=fedora-logo-icon
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```
 
gnome-shell-extension-pop-shell-0.1.0-0.2.20200821git8416328.fc32.noarch (built from commit 8416328 in [COPR](https://copr.fedorainfracloud.org/coprs/carlwgeorge/gnome-shell-extension-pop-shell/))

**Issue/Bug Description:**

Moving a window one monitor to the left or right via the keyboard shortcut does not work with tiling enabled.  The window does initially move as instructed, but immediately returns to the monitor it started on.  This is reproducible in a Pop!_OS live environment as well, so I don't believe it's a problem with the RPM packaging.  Here is a video of the behavior.

https://imgur.com/RpuI3xp

1. enable tiling mode
2. create one or more windows on the first monitor
3. press `<Shift><Super>Right` to move a window to the second monitor

**Expected behavior:**

window moves to second monitor and stays there

### LINKED ISSUE #538: Move app to another monitor unexpected behaviour
 ```NAME="Pop!_OS"
VERSION="20.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
VERSION_ID="20.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
LOGO=distributor-logo-pop-os
```
 
 
**Issue/Bug Description:**
Vertical screens are ignored when moving app from **_monitor to monitor_**. I think this could be realed to different resolutions on the monitors a the app keeps the previous monitor size when IMHO it should adjust to the new monitor resolution/scale

1. Setup 3 monitors
2. Set monitor layout as per attachement 1
3. Use shortcuts(Disable auto-tiling: I can't find  the related Issue where moving app from monitor to monitor does not work with auto-tiling ON):
image
4. See attachments for issue:

original_layout
0_original_layout
after_shortcut: `Shift + Ctrl + Super + Down`
1_after_shortcut
after_enabling_tiling
2_after_enabling_tiling

**Expected behavior:**
Move app to the monitor on direction to work on all monitor laytouts
3_expected

**Other Notes:**
I manage to move app to monitor after multiple iterations of `Super + Ctrl + Shift + H | J | K | L` but is not consistent

### PR DESCRIPTION
**Pull request by mmstick**
Merge with https://github.com/pop-os/session/pull/22

Fixes #538 
Fixes #522 
Closes #150
Closes #488

### DIFF
--- a/dark.css
+++ b/dark.css
@@ -38,3 +38,21 @@
     /* Seems to get the width just right to fit 3 digits */
     width: 75px;
 }
+
+.pop-shell-tab {
+    border: 1px solid #333;
+    color: #000;
+    padding: 0 1em;
+}
+
+.pop-shell-tab-active {
+    background: #FBB86C;
+}
+
+.pop-shell-tab-inactive {
+    background: #9B8E8A;
+}
+
+.pop-shell-tab-urgent {
+    background: #D00;
+}
--- a/light.css
+++ b/light.css
@@ -37,3 +37,21 @@
     /* Seems to get the width just right to fit 3 digits */
     width: 75px;
 }
+
+.pop-shell-tab {
+    border: 1px solid #333;
+    color: #000;
+    padding: 0 1em;
+}
+
+.pop-shell-tab-active {
+    background: #FFAD00;
+}
+
+.pop-shell-tab-inactive {
+    background: #9B8E8A;
+}
+
+.pop-shell-tab-urgent {
+    background: #D00;
+}
--- a/rebuild.sh
+++ b/rebuild.sh
@@ -48,13 +48,13 @@ dconf write ${KEYS_GNOME_WM}/move-to-monitor-down "@as []"
 
 # Super + direction keys, move window left and right monitors, or up and down workspaces
 # Move window one monitor to the left
-dconf write ${KEYS_GNOME_WM}/move-to-monitor-left "['<Shift><Super>Left','<Shift><Super>${left}']"
+dconf write ${KEYS_GNOME_WM}/move-to-monitor-left "@as []"
 # Move window one workspace down
-dconf write ${KEYS_GNOME_WM}/move-to-workspace-down "['<Shift><Super>Down','<Shift><Super>${down}']"
+dconf write ${KEYS_GNOME_WM}/move-to-workspace-down "@as []"
 # Move window one workspace up
-dconf write ${KEYS_GNOME_WM}/move-to-workspace-up "['<Shift><Super>Up','<Shift><Super>${up}']"
+dconf write ${KEYS_GNOME_WM}/move-to-workspace-up "@as []"
 # Move window one monitor to the right
-dconf write ${KEYS_GNOME_WM}/move-to-monitor-right "['<Shift><Super>Right','<Shift><Super>${right}']"
+dconf write ${KEYS_GNOME_WM}/move-to-monitor-right "@as []"
 
 # Super + Ctrl + direction keys, change workspaces, move focus between monitors
 # Move to workspace below
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -76,6 +76,16 @@
         </key>
 
         <!-- Window Management Keys -->
+        <key type="as" name="toggle-stacking">
+            <default><![CDATA[['s']]]></default>
+            <summary>Toggle stacking mode inside management mode</summary>
+        </key>
+
+        <key type="as" name="toggle-stacking-global">
+            <default><![CDATA[['<Super>s']]]></default>
+            <summary>Toggle stacking mode outside management mode</summary>
+        </key>
+
         <key type="as" name="management-orientation">
             <default><![CDATA[['o']]]></default>
             <summary>Toggle tiling orientation</summary>
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -184,5 +184,37 @@
             <default><![CDATA[['<Primary>Right','<Primary>l']]]></default>
             <summary>Swap window right</summary>
         </key>
+
+        <!-- Workspace Management -->
+
+        <key type="as" name="pop-workspace-down">
+            <default><![CDATA[['<Super><Shift>Down','<Super><Shift>j']]]></default>
+            <summary>Move window to the lower workspace</summary>
+        </key>
+
+        <key type="as" name="pop-workspace-up">
+            <default><![CDATA[['<Super><Shift>Up','<Super><Shift>k']]]></default>
+            <summary>Move window to the upper workspace</summary>
+        </key>
+
+        <key type="as" name="pop-monitor-down">
+            <default><![CDATA[['<Super><Shift><Primary>Down','<Super><Shift><Primary>j']]]></default>
+            <summary>Move window to the lower monitor</summary>
+        </key>
+
+        <key type="as" name="pop-monitor-up">
+            <default><![CDATA[['<Super><Shift><Primary>Up','<Super><Shift><Primary>k']]]></default>
+            <summary>Move window to the upper monitor</summary>
+        </key>
+
+        <key type="as" name="pop-monitor-left">
+            <default><![CDATA[['<Super><Shift>Left','<Super><Shift>h']]]></default>
+            <summary>Move window to the leftward monitor</summary>
+        </key>
+
+        <key type="as" name="pop-monitor-right">
+            <default><![CDATA[['<Super><Shift>Right','<Super><Shift>l']]]></default>
+            <summary>Move window to the rightward monitor</summary>
+        </key>
     </schema>
 </schemalist>
--- a/src/arena.ts
+++ b/src/arena.ts
@@ -1,0 +1,41 @@
+/** Hop slot arena allocator */
+export class Arena<T> {
+    private slots: Array<null | T> = new Array();
+    private unused: Array<number> = new Array()
+
+    truncate(n: number) {
+        this.slots.splice(n);
+        this.unused.splice(n);
+    }
+
+    get(n: number): null | T {
+        return this.slots[n];
+    }
+
+    insert(v: T): number {
+        let n;
+        const slot = this.unused.pop();
+        if (slot !== undefined) {
+            n = slot;
+            this.slots[n] = v;
+        } else {
+            n = this.slots.length;
+            this.slots.push(v);
+        }
+
+        return n;
+    }
+
+    remove(n: number): null | T {
+        const v = this.slots[n];
+        this.slots[n] = null;
+        this.unused.push(n);
+        return v;
+    }
+
+    * values(): IterableIterator<T> {
+        for (const v of this.slots) {
+            if (v !== null) yield v;
+        }
+    }
+}
--- a/src/arena.ts
+++ b/src/arena.ts
@@ -27,6 +27,7 @@ export class Arena<T> {
     }
 
     remove(n: number): null | T {
+        if (this.slots[n] === null) return null;
         const v = this.slots[n];
         this.slots[n] = null;
         this.unused.push(n);
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -207,8 +207,12 @@ export class AutoTiler {
     }
 
     /** Destroy all widgets owned by this object. Call before dropping. */
-    destroy() {
+    destroy(ext: Ext) {
         for (const stack of this.forest.stacks.values()) stack.destroy();
+
+        for (const window of ext.windows.values()) {
+            window.stack = null;
+        }
 
         this.forest.stacks.truncate(0);
     }
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -5,6 +5,7 @@ import * as lib from 'lib';
 import * as log from 'log';
 import * as node from 'node';
 import * as result from 'result';
+import * as stack from 'stack';
 
 import type { Entity } from 'ecs';
 import type { Ext } from 'extension';
@@ -14,6 +15,7 @@ import type { Rectangle } from 'rectangle';
 import type { Result } from 'result';
 import type { ShellWindow } from 'window';
 
+const { Stack } = stack;
 const { Ok, Err, ERR } = result;
 const { NodeKind } = node;
 const Tags = Me.imports.tags;
@@ -32,19 +34,63 @@ export class AutoTiler {
      * Call this when a window has swapped positions with another, so that we
      * may update the associations in the auto-tiler world.
      */
-    attach_swap(a: Entity, b: Entity) {
-        const a_ent = this.attached.remove(a);
-        const b_ent = this.attached.remove(b);
+    attach_swap(ext: Ext, a: Entity, b: Entity) {
+        const a_ent = this.attached.get(a),
+            b_ent = this.attached.get(b);
 
-        if (a_ent) {
-            this.forest.forks.with(a_ent, (fork) => fork.replace_window(a, b));
-            this.attached.insert(b, a_ent);
+        let a_win = ext.windows.get(a),
+            b_win = ext.windows.get(b);
+
+        if (!a_ent || !b_ent || !a_win || !b_win) return;
+
+        const a_fork = this.forest.forks.get(a_ent),
+            b_fork = this.forest.forks.get(b_ent);
+
+        if (!a_fork || !b_fork) return;
+
+        const a_stack = a_win.stack, b_stack = b_win.stack;
+
+        if (ext.auto_tiler) {
+            if (a_win.stack !== null) {
+                const stack = ext.auto_tiler.forest.stacks.get(a_win.stack);
+                if (stack) {
+                    a = stack.active;
+                    a_win = ext.windows.get(a);
+                    if (!a_win) return;
+
+                    stack.deactivate(a_win);
+                }
+            }
+
+            if (b_win.stack !== null) {
+                const stack = ext.auto_tiler.forest.stacks.get(b_win.stack);
+                if (stack) {
+                    b = stack.active;
+                    b_win = ext.windows.get(b);
+                    if (!b_win) return;
+
+                    stack.deactivate(b_win);
+                }
+            }
         }
 
-        if (b_ent) {
-            this.forest.forks.with(b_ent, (fork) => fork.replace_window(b, a));
-            this.attached.insert(a, b_ent);
-        }
+        const a_fn = a_fork.replace_window(ext, a_win, b_win);
+        this.attached.insert(b, a_ent);
+
+        const b_fn = b_fork.replace_window(ext, b_win, a_win);
+        this.attached.insert(a, b_ent);
+
+        if (a_fn) a_fn();
+        if (b_fn) b_fn();
+
+        a_win.stack = b_stack;
+        b_win.stack = a_stack;
+
+        a_win.meta.get_compositor_private()?.show();
+        b_win.meta.get_compositor_private()?.show();
+
+        this.tile(ext, a_fork, a_fork.area);
+        this.tile(ext, b_fork, b_fork.area);
     }
 
     update_toplevel(ext: Ext, fork: Fork, monitor: number, smart_gaps: boolean) {
@@ -57,8 +103,8 @@ export class AutoTiler {
             rect.height -= ext.gap_outer * 2;
         }
 
-        if (fork.left.kind === NodeKind.WINDOW) {
-            const win = ext.windows.get(fork.left.entity);
+        if (fork.left.inner.kind === 2) {
+            const win = ext.windows.get(fork.left.inner.entity);
             if (win) {
                 win.smart_gapped = smart_gaps && fork.right === null;
             }
@@ -97,12 +143,11 @@ export class AutoTiler {
         win.smart_gapped = smart_gaps;
 
         this.tile(ext, fork, rect);
-        this.log_tree_nodes(ext);
     }
 
     /** Tiles a window into another */
-    attach_to_window(ext: Ext, attachee: ShellWindow, attacher: ShellWindow, cursor: Rectangle) {
-        let attached = this.forest.attach_window(ext, attachee.entity, attacher.entity, cursor);
+    attach_to_window(ext: Ext, attachee: ShellWindow, attacher: ShellWindow, cursor: Rectangle, stack_from_left: boolean = true) {
+        let attached = this.forest.attach_window(ext, attachee.entity, attacher.entity, cursor, stack_from_left);
 
         if (attached) {
             const [, fork] = attached;
@@ -120,14 +165,11 @@ export class AutoTiler {
                 }
 
                 this.tile(ext, fork, fork.area.clone());
-                this.log_tree_nodes(ext);
                 return true;
             } else {
                 log.error(`missing monitor association for Window(${attachee.entity})`);
             }
         }
-
-        this.log_tree_nodes(ext);
     }
 
     /** Tile a window onto a workspace */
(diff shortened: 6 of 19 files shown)

### CODE REVIEW

**Comment by mmstick** [comment]
@jacobgkau Is there a good way to replicate this condition?

**Comment by jacobgkau** [comment]
> @jacobgkau Is there a good way to replicate this condition?

I haven't found an exact step-by-step process to reproduce it yet, but I did just get it to happen again on a fresh boot. I basically just opened up windows, created a stack, and started moving/swapping things around until it occurred. This time, I only had two displays attached.

shell-pr481-screenshot2-v2

(The red box is covering up Geary because my email is visible.) Pop Shop is what I'm focused on in the screenshot (and is hidden.) It took a little while to occur. I was initially trying to get it to happen with only a single display, I ended up attaching the second display because it didn't seem like I was getting anywhere with a single display (not sure if the second display is actually related, though.)

I'm trying to address some of the other shell PRs that are open, but when there's time, I can start a screen recording just after booting and try to get a recording of this happening if all else fails (that would include the exact steps taken to get to that point.)

**Comment by mmstick** [comment]
I'm working on it. I see that it happens because the actor gets hidden. Fix releasing soon.

**Comment by jmmaranan** [comment]
@mmstick - can easily reproduce it by toggling the opened windows (minimize/restore) in Dash to Dock. But don't think it is Dash to Dock. It does not need to be stacked previously. Am on stacking branch

image

**Comment by WatchMkr** [comment]
This is looking great. A few things:

1. Unable to resize a stack width with shortcuts or the mouse
2. When a stack is above another window that's not in a stack, I can only increase the stack height not decrease and the active hint remains at the original window location
3. When moving windows outside a stack, they often move to a location I don't expect. Left/right results in the window moving above or below the stack instead of to the left. Up/down moves the window to the left or right of the stack. It seems the logic is simply inverted.

**Comment by WatchMkr** [comment]
After leaving my laptop locked overnight, windows were all on one of two external displays and covering each other and the stacks. I'll try to get an easier way to reproduce this. Suspend and resume worked well.

**Comment by jacobgkau** [comment]
I'm still seeing some inpredictability when swapping windows in and out of stacks. Take a look at this example: https://www.dropbox.com/s/eyfugswr7oh2gnk/shell-pr481-screencast1.webm?dl=0 

Pop Shop was not selected or visible. Why did it come out of the stack instead of Firefox? Is this intentional because it's assumed I wouldn't swap the two windows that I'd just previously swapped?

**Comment by mmstick** [comment]
All seems to be well now. The issue with stack swapping was that a focus selection sometimes selected a hidden window in the stack.

**Comment by jacobgkau** [comment]
@mmstick I'm now seeing an issue with the active hint appearing over a stack instead of the focused window:

shell-pr481-screenshot3

In this example, Chromium is focused, but the active hint is appearing over the stack. If I focus on Gedit (next to the stack), the active hint appears over Gedit correctly, but moving down to either Chromium or the terminal results in the active hint going back to the stack. Here is a screen recording to better demonstrate: https://www.dropbox.com/s/dgtodbyuugwtzc2/shell-pr481-screencast2.webm?dl=0 

This is with `8cb674c`.

**Comment by mmstick** [comment]
@jacobgkau Is that the only issue you see? Trying to replicate it right now

**Comment by jacobgkau** [comment]
> Is that the only issue you see?

Yes, just completed the full test and I did not run into any other issues (that were not already present before this PR.) [shell-pr481.txt](https://github.com/pop-os/shell/files/5126562/shell-pr481.txt)

**Comment by WatchMkr** [comment]
I've seen a ghost window allocated when removing and adding windows from a stack today. 

Screenshot from 2020-08-25 16-06-11

**Comment by mmstick** [comment]
@WatchMkr Did that happen on moving the last window from the stack?

**Comment by WatchMkr** [comment]
Yes, I think that's when I'm seeing it.

**Comment by WatchMkr** [comment]
I also see the orientation change unexpectedly when I remove the some windows from a stack.

https://www.dropbox.com/s/ts5p8mcldwgz9ow/Screencast%20from%2008-25-2020%2004%3A30%3A03%20PM.webm?dl=0

**Comment by jacobgkau** [comment]
On `28ee72b`, I am still able to recreate the ghost window after moving the last window out of a stack: https://www.dropbox.com/s/d0a9gccd5u8rh8x/shell-pr481-screencast3.webm?dl=0

**Comment by jmmaranan** [comment]
If the change on `28ee72b` is applied it shows that issue using the keyboard to move the windows out of the stack - I think is the best way to replicate it. Mouse dragging not too much.

**Comment by mmstick** [comment]
Finally found the cause of the missing forks on detach. So this should be ready now.

**Comment by jmmaranan** [comment]
@mmstick - there is one more.

To repro:

- Tile 4 windows in monitor 1.
- Have 0 or more windows in monitor 2
- The lower left should have 2 windows. Stack both of them. Remove by mouse and drag to either the right fork or monitor 2.
- Drag the top-left or now left window that occupied the left side.
- Leaving empty space on left and doesnt retile. Keyboard or mouse movement.

**Comment by jacobgkau** [comment]
I was able to recreate the ghost window on `6b8047d`:

shell-pr481-screenshot5

It took an extra step or two from what @jmmaranan described, but I was able to get there by creating a stack and removing the last window from the stack a couple of times.

**Comment by mmstick** [comment]
Found the cause. I'll have a fix soon

**Comment by jmmaranan** [comment]
@mmstick - looks like it is working now after that change. I found #521 but can also reproduce on 841632872d77abc3c45cd13540ffae755839655a (master_focal)

**Comment by jmmaranan** [comment]
@mmstick, another issue, see if you can reproduce:
- Tile two windows, and enabled stack for each window.
- Move one of them down one workspace. Then move it back up to the same workspace as the other stacked one.
- The moved window does not attach to the stack. Not sure if that was intended. However, I can the tab now floating.
- The other window still on the stack, can be dragged but the stack tab would follow it, and then eventually it will detach to the window and leaving the tab somewhere with a space.

**Comment by jmbuhr** [comment]
Sometimes the opposite of a ghost window also happens, windows are able to cover each other up even though they should be tiling.
To reproduce, open 2 windows and make them both their own stack, then add another window to each stack. After moving all windows from the stacks (using window management mode) and moving them around with Super + Mouse a bit, they start to irreversibly cover each other up at some point. Here, what looks like two windows:

image

Is actually 4:

image

**Comment by jmmaranan** [comment]
@jmbuhr - did the windows happen to start in different workspaces? It looks like #521 and it happens on master too.

**Comment by jmbuhr** [comment]
I have observed #521 as well, but wasn't able to create the bug reproducibly until this method. The windows all came from the same workspace, just pressing Super + T 4 times, then stacking and unstacking the windows. The ghost windows and overlapping windows are probably distinct but similar issues.

**Comment by jacobgkau** [comment]
I'm seeing larger breakage when swapping windows inside and outside of a stack on `19d76c5`: https://www.dropbox.com/s/ot0dvcb3bml4ipt/shell-pr481-screencast4.webm?dl=0 

The problem seems to occur when I move a window out of a stack, then attempt to swap that window with one of the windows in the stack. The windows don't actually swap, but the tabs indicate that Shell thinks they swapped, and then I get ghost windows, disappearing tabs, etc.

Here are two more examples. Vertical: https://www.dropbox.com/s/kb5u6vbexiq1jtn/shell-pr481-screencast5.webm?dl=0 

Horizontal: https://www.dropbox.com/s/a94vd7quazeiufs/shell-pr481-screencast6.webm?dl=0

**Comment by mmstick** [comment]
I guess that last commit breaks more things than it fixes

**Comment by jmmaranan** [comment]
@mmstick - I think it is an issue before the last change that was reverted. It might be an issue still, will defer to @jacobgkau though. The change that was reverted was for when switching workspaces and cleanly prevent workspace switch from throwing js errors on the entity. But will defer to QA.

**Comment by mmstick** [comment]
I did experience the same issues on that commit, and reverting it fixed it. I have found another solution for the lack of entity, and I'm currently working on windows moving back to a workspace with one stack.

**Comment by mmstick** [comment]
Okay. I think I've got it now. Movements between workspaces are working. Largely caused by a window detached from a stack via the keyboard having the incorrect parent association.

**Comment by jmbuhr** [comment]
I can confirm that the previous issue I observed (https://github.com/pop-os/shell/pull/481#issuecomment-683710566) is no longer present after the commit.

**Comment by jmbuhr** [comment]
One interface question: Using Super + S creates a stack from one Tile and removes the stack if it only has one window in it. Shouldn't it also dissolve a whole stack of multiple windows back into tiles?

**Comment by mmstick** [comment]
@jmbuhr Wouldn't want to have an accidental use of `Super + S` throwing windows around the tree. It's safe to unstack when there's only one window, but it's ignored if there's more than one.

**Comment by mmstick** [comment]
I guess it now comes down to reviews from @maria-komarova  and @WatchMkr before we proceed with merging this feature.

**Comment by mmstick** [comment]
I found one issue, but it's probably an easy fix. If you destroy a window that's in a stack, there's a reference to it still hanging around in GNOME Shell, since it remains in the dash.

**Comment by mmstick** [comment]
Finally found the cause of the applications remaining in the dash.
That was rather annoying to track down.

**Comment by jmbuhr** [comment]
I found another weird edge case. With the default dynamic workspaces enabled:
- Open 1 window on the first workspace.
- Move to the second workspace
- Open 2 windows in a stack
- Move to workspace 1 and close the lonely window
- Trigger the dynamic workspace re-numbering by moving to workspace 2

The 2 windows are now no longer stacked and sort of invisible. They can only be seen in the Super key window overview. The minimize shortcut and then clicking them in the overview brought them back but the gnome dash icons are confused and need a alt+f2+r reset.
I replicated it before and after the latest commits.

**Comment by mmstick** [comment]
I think I know what's happening there

**Comment by mmstick** [comment]
Okay, that should be fixed now. Needed to update the workspace number of stacks when its workspace changes

**Comment by leviport** [comment]
I think I'm seeing that same un-stacking bug if I have a stack of two windows next to a single window, then I move that single window upwards to a new workspace with `Super`+`Shift`+`Up`.

Otherwise, it's looking real good and [passing regression testing](https://github.com/pop-os/shell/files/5176589/pop-shell-stacking-checklist.txt).

**Comment by jmmaranan** [comment]
@mmstick - when I tried @leviport's steps/setup, I am seeing a window unable to attach back to the stack if I swap it back and forth between workspaces quickly. But if _gently_ Super + Shift + Up/Down, it would attach nicely.

**Comment by jacobgkau** [comment]
I'm now seeing an issue on this branch with windows being slightly offset when un-maximized or un-fullscreened. This only occurs some of the time.

Here's an example with Firefox: https://www.dropbox.com/s/rbcfqv6mmqnt7wv/shell-pr481-screencast8.webm?dl=0 

The window pops back into the correct position when any tiling action is performed, including simply moving the focus using Super+direction keys. Here's a screenshot of this issue happening with the terminal:

shell-pr481-screenshot1

Sometimes, Shell seems to correct the position very quickly, like this: https://www.dropbox.com/s/aj9qm73mp0u6xfg/shell-pr481-screencast9.webm?dl=0 

I have not been able to reproduce this on stable. Stable seems to un-maximize much quicker, with less of an animation.

**Comment by jacobgkau** [comment]
Another issue, Super+O is not changing window orientation right now.

**Comment by mmstick** [comment]
The Super+O issue is also on master. Haven't found a way to reproduce it yet.

**Comment by mmstick** [comment]
I removed some retiling triggers so it's not excessively retiling multiple times in a row. I guess one of those retiles was fixing windows that unmaximize out of place

**Comment by jacobgkau** [comment]
Things are looking very good now, stacking and moving between displays are both working well: [shell-pr481-run7.txt](https://github.com/pop-os/shell/files/5228530/shell-pr481-run7.txt) I can confirm this fixes https://github.com/pop-os/shell/issues/522 and https://github.com/pop-os/shell/issues/538.

I did discover that if a floated window is moved between workspaces using the keyboard, any stacking tabs end up overlapping the focused floated window:

shell-pr481-screenshot2

Moving the floated window doesn't bring it back to the top, but changing focus does bring it back to the top.

Finally, a floated window's position is reset to the top left of the display when it's moved between workspaces with the keyboard. This seems like a regression.

**Comment by jmbuhr** [comment]
I noticed one additional thing:

When having multiple windows on the first workspace, I used to be able to move one window up a workspace (creating a new first workspace) because the workspace is dynamically created. This no longer works (I believe as of https://github.com/pop-os/shell/pull/481/commits/029a99ecb37644846c9f28bd053b88ec8d58c717 ).

**Comment by mmstick** [comment]
@jmbuhr Try again. I've just implemented support for that.

**Comment by mmstick** [comment]
@jacobgkau All seems to be in order now.

**Comment by jmbuhr** [comment]
After the lastest commits, when I have 2 windows on one monitor and move one of them to another monitor, the other follows. And after some more movement they loose track of which tree they belong to and no longer tile.

**Comment by mmstick** [comment]
@jmbuhr I'll have this fixed within the hour. I think it's just not setting the correct workspace number after windows are shifted down a workspace.

**Comment by mmstick** [comment]
@jmbuhr Does this only happen when moving windows with the mouse?

**Comment by jmbuhr** [comment]
Yes. The Ctrl+Shift+Super+Direction Shortcut works as expected. But as soon as a window is moved with the mouse from a monitor with 2 windows onto a different monitor, the remaining window feels lonely and follows. This does not happen with 3 or more windows.

**Comment by mmstick** [comment]
@jmbuhr Do you have exact steps for getting this to happen?

**Comment by jmbuhr** [comment]
Sure. Setup: I am on the latest pop-os, and the latest commit of this branch. I have a screen set up above my laptop screen.

In Tiling Mode, open any two windows e.g. press Super+B oder Super+T twice, grab one of them with Super+Mouseclick and drag it to the other screen (doesn't matter which direction). The other window follows and they both occupy the whole screen.

**Comment by jmmaranan** [comment]
@jmbuhr - did you have the smart gaps turned ON?

**Comment by jmbuhr** [comment]
o.O Yes! And turning them off makes the issue go away.

**Comment by jmmaranan** [comment]
Thanks @jmbuhr. @mmstick, that's an issue on master. It is somewhere on the detach window smart gap flow.

**Comment by mmstick** [comment]
Excellent. Very helpful. I've been looking in the wrong places

**Comment by jacobgkau** [comment]
@mmstick The active hint is no longer appearing over stacks. Reviewing my earlier screenshots, this was not the case previously. Was this changed intentionally? I thought you were planning to rebase stacking to include the active hint improvements in https://github.com/pop-os/shell/pull/544, but that doesn't seem to be the case either.

**Comment by mmstick** [comment]
This branch includes #544. Didn't notice that it's not working anymore. Probably caused by incompatible changes in the rebase

**Comment by jacobgkau** [comment]
One reason I thought it didn't include #544 is because maximized windows are having their hints hidden, while hints should display on maximized windows after #544. I did confirm that I'm running `5476e69` on the machines I'm testing with now.

**Comment by mmstick** [comment]
The commit that adds the stacking feature is breaking it. Not yet sure why because all the code from #544 made it into the rebase. I'll find the fix soon.

**Comment by jmmaranan** [comment]
@mmstick - stack.ts is missing the new css class name. `pop-shell-border-normal`
```
border: St.Bin = new St.Bin({ style_class: 'pop-shell-active-hint pop-shell-border-normal' });
```

jacobgkau:
I'm seeing some instances again where a stack incorrectly gets the active hint when certain non-stacked windows are selected: https://www.dropbox.com/s/l9owcc2e7yv13fk/shell-pr481-screencast10.webm?dl=0 

I'm able to reproduce this by:
- Stacking some windows
- Turning tiling off and then back on
- Re-stacking some of the windows that were previously stacked, but not all
- Selecting one of the windows that was in the stack before I toggled tiling off and back on, and is no longer stacked

In addition to that issue, I also noticed that smart-gapped windows have an active hint on a secondary display, but not the primary display (it's hanging off the side of the window on the primary display): https://www.dropbox.com/s/bw68qquproqk731/shell-pr481-screencast11.webm?dl=0 

Maximized windows still aren't getting active hints, is that expected or are we expecting them to get hints again?

mmstick:
I'm still working on the active hint issues

jmmaranan:
I added #558 - for the tiling toggle when stacked, active hint issue, the window.ts line 332, need to clean up the trace of the stack. Because stack === null never evaluates to true. Or just remove it but when tapping the actual window on focus it shows the border on both stack tab and the windows.

jmmaranan:
@jacobgkau the smart gap hint lying outside the primary is a master issue.

mmstick:
I think we're in a good state now. I've been doing a lot of testing today with multiple displays, and everything seems to be in order with these last few commits.

jacobgkau:
Moving a window to a new first/top workspace un-stacks windows on other workspaces: https://www.dropbox.com/s/8qg8mahbldfmprl/shell-pr481-screencast12.webm?dl=0 

The window I'm moving doesn't need to have been part of the stack for this to occur.

mmstick:
@jacobgkau If that was with the mouse, there's nothing I can do to fix that since GNOME Shell doesn't tell us this is happening.

jacobgkau:
> @jacobgkau If that was with the mouse, there's nothing I can do to fix that since GNOME Shell doesn't tell us this is happening.

Yes, it was with the mouse. Understood, I'll note it as expected behavior and move on.

mmstick:
I found an issue. One of the recent commits is causing workspace movements upwards to render windows to the wrong workspace

mmstick:
Okay. I found the issue. Stack windows just weren't being moved to their newly-assigned workspaces.

mmstick:
@jacobgkau Okay, I think this is ready for a second look

jacobgkau:
[shell-pr481-run11.txt](https://github.com/pop-os/shell/files/5271453/shell-pr481-run11.txt) It looks almost perfect. The only thing I noticed this time around is that maximized non-stacked windows have an active hint, while maximized stacked windows do not have a hint: https://www.dropbox.com/s/443oez4tqbja1p0/shell-pr481-screencast13.webm?dl=0

jmmaranan:
I can help with that the stacked maximize. Will send a PR later

jmmaranan:
@jacobgkau @mmstick - please bear with me. I have to redirect all hint logic to the active window so we don't duplicate some of the logic that was done in master. Looking at the implementation with the stack, I am voting to have the window as still the primary actor even within the stack, and the stack is a decoration, since there is only one window focused at a time.

I am finishing up on the grab_op. Should send PR soon.

jmmaranan:
Fix for the maximized hint on stack or not: https://github.com/pop-os/shell/pull/562

mmstick:
With that last commit , maximized stacked windows have the hint now.

jacobgkau:
The tab bar is sometimes covering up Firefox if it was stacked and then full-screened:

shell-pr481-screenshot6

It seems to not happen the first time it's full-screened after being selected in the stack, but happens subsequent times that it's full-screened until I navigate to another item in the stack and back. Also affects Chromium (and presumably any full-screenable apps.)

jmmaranan:
@jacobgkau - it is happening in 3b21d271467769037f0415cf8b853e010d361a3f. So it is not introduced in the 2 recent commits. I will check

jacobgkau:
Another thing I noticed, when moving a window across displays, I'm able to make it cover up the rest of the windows on the second display if I try moving it past the edge of the display. https://www.dropbox.com/s/fp1z1p5tgmzi99i/shell-pr481-screencast14.webm?dl=0 

Here's another example with less windows open and no stacks: https://www.dropbox.com/s/ayylv7ucaei0x88/shell-pr481-screencast15.webm?dl=0 

It's almost like the window is being floated, except it's not, I'm getting to that point just by pressing the arrow keys in window management mode, and I'm able to move the window back into a proper position with more arrow keys. Sometimes, this only happens once and the window doesn't cover up other windows again once it's been placed down, but other times, I'm able to move it back and forth between the larger state and the correct state.

jmmaranan:
@mmstick - added fix for hiding tabs on full screen.

jmmaranan:
@jacobgkau - for the last one, I'd probably leave that to @mmstick

mmstick:
I'm looking into that last one

mmstick:
Fixed the tree creation on trying to move a window off a display with management mode movements

jacobgkau:
I've got a general observation and an issue that seems related. Right now, turning auto-tiling off and back on sometimes moves all of the windows back to the primary display. I'm used to seeing windows get moved around, but I don't recall ever seeing the secondary display get cleared out like this. https://www.dropbox.com/s/1t34k9aiu8w5rhx/shell-pr481-screencast16.webm?dl=0 

The issue is that if I have windows on more than one workspace and I turn auto-tiling off and back on, the windows on the non-visible workspace get restricted to a certain area. At the same time, the shell is reserving space for the non-visible windows on visible workspace. https://www.dropbox.com/s/aowknacmck8cme4/shell-pr481-screencast17.webm?dl=0 

If I move windows around to create more ghost window space, the windows on the other workspace are resized to fill that empty space. https://www.dropbox.com/s/ry849qz3mi5xfow/shell-pr481-screencast18.webm?dl=0

mmstick:
@jacobgkau Does that still happen with `4fc8df1f50c28cddbac31a88dd3647eb9bd6452c`?

jacobgkau:
> Does that still happen with `4fc8df1f50c28cddbac31a88dd3647eb9bd6452c`?

No, after reverting to 4fc8df1f50c28cddbac31a88dd3647eb9bd6452c I am not seeing either of those behaviors.

mmstick:
Okay, I amended that last commit so it won't do that anymore

mmstick:
Fixed the active hint border still being around what used to be a stack after disabling tiling

jacobgkau (approved):
All regression testing is passing and I am not seeing any bugs: [shell-pr481-run15.txt](https://github.com/pop-os/shell/files/5279565/shell-pr481-run15.txt)

One final observation, it's sometimes unintuitive which window is focused upon switching focus to a stack just after swapping a window in the stack with a window outside of the stack. In some cases, instead of the visible window (which was just swapped into the stack) being focused, the previously-selected window in the stack is focused.

Vertical example: https://www.dropbox.com/s/nsu1plsqtp3nzip/shell-pr481-screencast19.webm?dl=0 

Horizontal example (which demonstrates that the order of focus matters): https://www.dropbox.com/s/otjy9fxwv1ry6h7/shell-pr481-screencast20.webm?dl=0 

This doesn't seem like a showstopper, it just took me a while to understand the logic. I can review again if this is something UX thinks should change.

jmmaranan:
@jacobgkau - can you check make listen when you combine in the stack or close windows? I am seeing an excessive polling in the logs. And the glib timeout never expires the first time it is called. I am working on a fix now.

jmmaranan:
https://github.com/pop-os/shell/pull/565 - this fixes the excessive polling.

jacobgkau (approved):

jackpot51 (approved):

PR #483, opened 2020-08-06
feat: Add configuration file for user-defined float list

### LINKED ISSUE #254: Application Always Float
One nice feature of i3 and other tiling window managers is the ability to make an application always float, or at least open floating.

Currently programs that typically start with a smaller window, such as the Calculator application that ships with Pop, get resized to an awkwardly large size to be tiled.

A similar issue, small dialog windows tend to get their own tile rather than floating over the window that spawned them.

Is there a way to make certain applications, or window titles, always float?
Can we get floating dialogs?

Also, is there a config file that i can edit and commit to my dotfiles?

Thanks for the hard work.

mmstick:
Closes #254

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -33,6 +33,11 @@
             <summary>Size of a row in the display grid</summary>
         </key>
 
+        <key type="b" name="smart-gaps">
+            <default>false</default>
+            <summary>Hide the outer gap when a tree contains only one window</summary>
+        </key>
+
         <key type="b" name="snap-to-grid">
             <default>false</default>
             <summary>Snaps windows to the tiling grid on drop</summary>
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -47,12 +47,17 @@ export class AutoTiler {
         }
     }
 
-    update_toplevel(ext: Ext, fork: Fork, monitor: number) {
+    update_toplevel(ext: Ext, fork: Fork, monitor: number, smart_gaps: boolean) {
         let rect = ext.monitor_work_area(monitor);
-        rect.x += ext.gap_outer;
-        rect.y += ext.gap_outer;
-        rect.width -= ext.gap_outer * 2;
-        rect.height -= ext.gap_outer * 2;
+
+        if (!(smart_gaps && fork.right === null)) {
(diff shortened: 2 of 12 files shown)


jmmaranan:
@mmstick wondering if this can be a string blob in gsettings?

mmstick:
I think this is too complex to be a string in gsettings. The floating list is an array of objects, and gsettings doesn't provide a text editor for editing values.

Since this extension is written in JS, it'd be easier for us to serialize and deserialize a JSON configuration file. A text file easily edited by a text editor that can provide syntax highlighting will be a more suitable choice than a registry.

We can watch the config for changes and dynamically reload and apply these values.

jacobgkau:
Any invalid JSON in the config file currently seems to stop Pop Shell from loading altogether. Is it possible to just ignore the config file if it's invalid? I do see that we log a `JS ERROR` line.

When I run `xprop` on the Calculator app, I see this line:

```
WM_CLASS(STRING) = "gnome-calculator", "Gnome-calculator"
```

Adding `"Class": "Gnome-calculator"` to the config file works, while `"Class": "gnome-calculator"` does not. Is this intended/is there an explanation for why only the latter version works?

https://github.com/pop-os/shell/issues/254 mentions adding toggles to GNOME Settings > Applications to make it easier to manage which applications float. Since we're closing the issue with this PR, are we not planning to add a GUI for this feature? (Even something like right-click titlebar -> Always Float so the user doesn't need to run `xprop` manually?)

mmstick:
@jacobgkau 

- I've wrapped the parsing in a try-catch, so it will return the default Config values on error.
- The string for the window class for GNOME Terminal that GNOME Shell gives us is `Gnome-calculator`.
- And unless you specify your regular expression to be case insensitive, case matters
- You can see the title and class of a window that GNOME sees by enabling `log_on_focus` and watching the gnome-shell logs. when you click on a window. We don't get these values from xprop, since you may not necessarily be on X11.
- We could add a toggle for gnome-control-center. A feature request could be created on that repo.

mmstick:
I think we could make a feature request here for an "Always Float" option. 🤔

mmstick:
Actually, it would be difficult to add GUI support for GNOME Control Center in the way described, because windows don't have any info about the application ID they were spawned from, and they're not guaranteed to have the same class or title on all of their windows.

mmstick:
There's no reason for the difference in field names for the logs.

PR #484, opened 2020-08-06
feat: Add smart gaps feature

### LINKED ISSUE #470: [Feature Request] Smart-gaps
A behaviour similar to smart-gaps in i3-gaps which when there's only one window/program open in a workspace, it fills the whole space without any gaps, when another program/window is started then the gaps are shown.

**Expected behaviour**:
1. If there's only one program in the workspace, the program's window is in maximized mode
2. When there's any other program, the gaps appears.

mmstick:
Which shall be disabled by default, but is a common request

Closes #470

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -33,6 +33,11 @@
             <summary>Size of a row in the display grid</summary>
         </key>
 
+        <key type="b" name="smart-gaps">
+            <default>false</default>
+            <summary>Hide the outer gap when a tree contains only one window</summary>
+        </key>
+
         <key type="b" name="snap-to-grid">
             <default>false</default>
             <summary>Snaps windows to the tiling grid on drop</summary>
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -47,12 +47,17 @@ export class AutoTiler {
         }
     }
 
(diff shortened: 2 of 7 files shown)


mmstick:
@jacobgkau Fixed now

jacobgkau:
@mmstick Is what's shown in this video expected behavior? https://www.dropbox.com/s/h3hdigjn5qsjnee/shell-pr484-video1.mkv?dl=0 

Based on #470 and the preference description, I was expecting smart gaps to only kick in if a window is the only window on a workspace/display. If the intention is to never have gaps on the outer edge, then I'm not understanding why moving the terminal between these two positions would cause the gaps to appear and disappear.

mmstick:
@jacobgkau I found the problem

leviport:
@mmstick did that last commit need to be tested before merging?

laichiaheng:
I still see the gap with only one window.

mmstick:
Is the AUR building from the master branch?

laichiaheng:
@mmstick It is from master_focal

mmstick:
It works here, so either you haven't restarted the shell, or it didn't build the latest commit.

laichiaheng:
> It works here, so either you haven't restarted the shell, or it didn't build the latest commit.

Nope. I have re-installed it and restart the shell many times.

mmstick:
It's working in Pop!_OS

Screenshot from 2020-08-11 20-43-14

laichiaheng:
> It's working in Pop!_OS
> 
> Screenshot from 2020-08-11 20-43-14

It appears now.

Issue #493, opened 2020-08-12
Active hint moves away from active floating window when tiling actions are performed


jacobgkau:
**Distribution (run `cat /etc/os-release`):** Pop!_OS 20.04 LTS
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):** Occurs in `0.1.0~1597190322~20.04~149061a` (current master) as well as `0.1.0~1597190879~20.04~52ded06`.

**Issue/Bug Description:** After floating a window, the active hint usually moves back to a window that is not currently active. Additionally, if a window is closed and causes other tiled windows to be adjusted, and the focus is placed on a floating window, the active hint ends up going to one of the adjusted windows instead of the active floating window.

**Steps to reproduce (if you know):**
- Scenario 1:
  - Enable active hints.
  - Tile two windows.
  - Float one of the windows with Super-G.
- Scenario 2:
  - Arrange four windows in a tree, then float one of them.
  - Select the floated window, then one of the tiled windows.
  - Close the tiled window with Super-Q.

**Expected behavior:** The active hint should always surround the active window, even if the active window is floated.

**Other Notes:**
- Video demonstrating both scenarios: https://www.dropbox.com/s/eyjx08s6b9zhpcu/active-hint-issue.webm?dl=0

jmmaranan:
@jacobgkau - can this be closed? Implemented in #503

PR #496, opened 2020-08-13
fix: Hide active hint when the last window in a tree is smart-gapped

mmstick:


jacobgkau:
This doesn't seem to be working in `7c21486`. Like I said when I mentioned this in https://github.com/pop-os/shell/pull/484, even with regular maximization, the active hint re-appears after focus is moved to another window and then back to the maximized window. When a window becomes smart-gapped because another window has been moved off the screen, the active window is the window the user moved off the screen, and as soon as focus is moved back to the smart-gapped window, the active hint reappears:

shell-pr496-screenshot1

(In this screenshot, I just moved the terminal from the primary display to the secondary display, then used Super-Left to focus back on Files.)

Even if I have Files by itself in a tree, toggle Smart Gaps off, make sure Files is focused, and toggle Smart Gaps back on, the active hint moves to the Pop Shell preferences window for a moment, then _incorrectly_ moves back to the non-active Files window (I think this is related to https://github.com/pop-os/shell/issues/493.)

shell-pr496-screenshot2

I'm not seeing this PR provide any benefit unless we provide a keyboard shortcut for toggling smart gaps (since that would allow the user to toggle smart gaps without taking focus away from the window being smart-gapped, assuming that results in the smart gap being hidden, which I don't know of a way to test.) Are we planning to fix the maximize active hint so it no longer shows up after un-focusing/re-focusing, and is it possible that would achieve this as well? (Alternatively, if the active hint is supposed to show up for maximized windows, then it should also show up when a window is initially maximized, without needing to re-focus.) Please let me know if I'm missing something.

PR #497, opened 2020-08-13
fix: Ability to snap to bottom edge of work area

### LINKED ISSUE #45: Grid leaves empty area at bottom of screen
This seems like it may be related to https://github.com/pop-os/shell/issues/39

If I use <kbd>Super</kbd> + <kbd>Enter</kbd> to resize a window while in free-float mode, there is a region on the bottom of the screen that I cannot resize the window into with <kbd>Shift</kbd> + arrows. I can use the mouse to expand the window into that area, but it's almost like the grid does not extend that far.

Screenshot from 2020-03-02 09-34-37

The screenshot above shows my monitor setup with an Alacritty window on each display sized as large as I can make it in window management mode. The top display is 4K (100% scaling) and the bottom one is 1080p. The empty area is much larger on the 4K display in terms of pixels, but it seems roughly the same proportionally as the 1080p display.

Also this does not seem to happen with tiling mode activated, so that tiling grid seems to line up with the full area of my displays better.

### LINKED ISSUE #463: Bottom of screen is unavailable in stacking mode
**Distribution (run `cat /etc/os-release`):** Fedora 32 (Workstation Edition)
 
 
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):** `9e8ab29cb976b078aa6e8fab59b09527a092a1b8`

**Issue/Bug Description:**
I've seen a few people post about this issue specifically for their terminals, but this is affecting every window for me. If I move/resize a window in stacking mode, there's a space at the bottom of the screen that the window can never move into. The space is about half the size of the row increment when resizing/moving a window, so I suspect this is to do with my screen resolution / DPI. Maybe if the amount of space left on the screen is less than a full row, it won't move into it?

Here you can see an example - the windows on the left have been arranged in auto-tile mode, which has no problem utilising the full screen. I then entered stacking mode and resized the right-hand window:
image

This is on a 1920x1080, 15.6" monitor
1. Enter stacking mode
2. Enter adjustment mode
3. Attempt to resize or move a window so it touches the bottom of the screen

**Expected behavior:**
The window touches the bottom of the screen, like it does in auto-tiling mode.

mmstick:
Closes #45
Closes #463

### DIFF
--- a/src/tiling.ts
+++ b/src/tiling.ts
@@ -63,8 +63,8 @@ export class Tiler {
     rect(ext: Ext, monitor: Rectangle): Rectangle | null {
         if (!ext.overlay.visible) return null;
 
-        const columns = monitor.width / ext.column_size;
-        const rows = monitor.height / ext.row_size;
+        const columns = Math.floor(monitor.width / ext.column_size);
+        const rows = Math.floor(monitor.height / ext.row_size);
 
         return monitor_rect(monitor, columns, rows);
     }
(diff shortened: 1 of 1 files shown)


mmstick:
Let's create an issue for that one

PR #499, opened 2020-08-13
fix(hint): hide hint on max refocus

jmmaranan:
This only fixes the maximized window being re-focused, no hints. 

The global.display knows which window is really being focused - that's why the active hint on minimized was fixed on #487 but seemed to introduce #493.

But the tiler is moving windows and the display emits focus events on float during that time.

### DIFF
--- a/src/active_hint.ts
+++ b/src/active_hint.ts
@@ -66,6 +66,7 @@ export class ActiveHint {
     }
 
     restack(actor: Clutter.Actor) {
+        this.hide();
         if (this.tracked) {
             if (this.tracked.workspace === global.workspace_manager.get_active_workspace_index()) {
                 // Avoid restacking if the window is maximized / fullscreen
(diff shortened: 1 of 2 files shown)


mmstick:
Will this be deprecated by #503 ?

jmmaranan:
#503 will have to be reapplied/adjusted to smart gaps if it isn't working.

Is smart gap by definition a maximized state for if only a single window?

jacobgkau:
> Will this be deprecated by #503 ?

#503 does fix manually maximized windows not getting an active hint after focusing on another window and then back to the maximized window (this is the correct behavior), so I would say this (#499) is deprecated by #503.

> Is smart gap by definition a maximized state for if only a single window?

I'd have to defer on how it's implemented, but I would define a smart gapped window as a single window in a tree that the shell makes _appear_ maximized by making it the maximum possible size with no outer gap (it seems like there is still a difference under the hood when I press Super-M to manually maximize a window, which is why I refer to that as a "manually maximized window.")

#503 causes an active hint to appear on top of a smart gapped window, which is a regression. When Smart Gaps are enabled on #503, I can basically toggle the active hint on and off of the window by pressing Super-M (which does not move or resize the window at all in that test case.)

jmmaranan:
@jacobgkau - that makes sense to still have the hint on a smart-gapped window. 

@mmstick - I have sort of `fixed` the issue on #503 when moving between workspaces. Well, the current implementation border is wider than the window. The restack did not necessarily moved the border at the top even if I had bounded every window/workspace signal there is :) . You will see when you tested it.

mmstick:
Closing in favor of #503

PR #503, opened 2020-08-16
fix(active-hint): re-implement active-hint

### LINKED ISSUE #493: Active hint moves away from active floating window when tiling actions are performed

jmmaranan:
Simplified tracking of border for each window using St.Bin.

Closes #493. #487 looks ok


src/window.ts
```diff
@@ -262,6 +272,46 @@ export class ShellWindow {
...
+                if (this.is_maximized() || this.meta.minimized) {
+                    this.hide_border();
+                }
+                this.update_border_layout();
+            }),
+            this.meta.connect('position-changed', () => {
+                this.update_border_layout();
+            }),
+        ];
+    }
```
mmstick, line 314: Would this work here? https://github.com/pop-os/shell/blob/master_focal/src/extension.ts#L381:L386

src/extension.ts
```diff
@@ -449,37 +443,37 @@ export class Ext extends Ecs.System<ExtEvent> {
...
+        //     let cws = this.workspace_id(null);
+
+        //     if (aws[0] === cws[0] && aws[1] === cws[1]) {
+        //         this.active_hint.track(active);
+        //     } else {
+        //         this.active_hint.untrack();
+        //     }
+        // };
+
+        // refocus_hint();
```
mmstick, line 476: We can remove all code that has been commented out.

src/window.ts
jmmaranan: It does not seem to work when I bounded the handler. That's why I added two new sets.

mmstick: Then we should definitely add these signals somewhere so we can disconnect them on extension disabling / window destruction.

src/extension.ts
jmmaranan: Removed all outdated code.

src/window.ts
jmmaranan: @mmstick - added on `this.ext.window_signals` push on location of binding

jmmaranan:
@mmstick - please do not merge this yet. I found an issue when single window and move workspace down. Or single window open, move it to another workspace. The active hint was not moving.

Let me see if it was the previous `refocus_hint()` on active workspace changed.

jacobgkau:
Expanding on part of my comment from #499, here is what a smart gapped window looks like on this branch:

Screenshot from 2020-08-18 15-58-13

And here is what a manually maximized (Super-M) window looks like:

Screenshot from 2020-08-18 15-58-18

I personally think this makes sense. The testing criteria includes the active hint being hidden for a maximized window, but I can imagine a user wanting the active hint to appear while they're using smart gaps-- with multiple displays, it might be unclear what's active if the hint is hidden. This behavior still allows them to hide the active hint for maximized windows manually.

I do see the issue with the active hint from a single window remaining when moving to a new workspace:

Screenshot from 2020-08-18 15-58-22

I'll wait for that to be addressed.

jmmaranan:
@jacobgkau, @mmstick - this now can be reviewed. 

One more discovery that a St.ThemeNode can be implemented to add feat to change colors of the border or sizes. Will add that up in a future PR.

jmmaranan:
@jacobgkau - added fix on need for border and window to be on same workspace when changing

jmmaranan:
It is currently using the window clone border class. I will add or update the existing css.

jmmaranan:
I am currently working on a way for users to customize it though. A limited color set and border size. We still have the stacking problem that's why thicker border becomes better when the window actor restacks at the top.

mmstick:
Thanks! All commits have been rebased to master

PR #508, opened 2020-08-19
Re-enable Wayland support

### LINKED ISSUE #166: Pop Shell makes Gnome unresponsive after login [Appears to be Wayland only]

mmstick:
It seems to be working now, although I had to disable the active hint feature since it causes issues. The window title support also does not work in Wayland since we don't have an xprop equivalent.

Closes #166

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1130,7 +1130,7 @@ export class Ext extends Ecs.System<ExtEvent> {
 
         // Modes
 
-        if (this.settings.tile_by_default() && !this.auto_tiler && !utils.is_wayland()) {
+        if (this.settings.tile_by_default() && !this.auto_tiler) {
             Log.info(`tile by default enabled`);
 
             this.auto_tiler = new auto_tiler.AutoTiler(
@@ -1213,8 +1213,6 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     toggle_tiling() {
-        if (utils.is_wayland()) return;
-
         if (this.auto_tiler) {
             Log.info(`tile by default disabled!`);
             this.unregister_storage(this.auto_tiler.attached);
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -435,7 +435,7 @@ export class Ext extends Ecs.System<ExtEvent> {
         this.column_size = this.settings.column_size() * this.dpi;
         this.row_size = this.settings.row_size() * this.dpi;
 
-        if (this.settings.active_hint() && !this.active_hint) {
+        if (this.settings.active_hint() && !this.active_hint && utils.is_wayland()) {
```
ids1024, line 438: Should this be `!utils.is_wayland()`?

(I'm not really familiar with exactly what this code is supposed to be doing, but I assume it shouldn't be changing behavior on X).

mmstick: You're right. This feature should be disabled for Wayland, for now.

mabachel:
> * In Wayland, some windows do not return to the proper position when un-maximizing with the keyboard. If I enter adjustment mode (Super+Enter), the adjustment overlay starts off where the window was supposed to be; if I resize other windows, that also makes the affected window return to its proper position. This only happens when tiling is enabled. Video to demonstrate: https://www.dropbox.com/s/3y2iftyzo8jv8ri/Screencast%20from%2008-20-2020%2005%3A37%3A52%20PM.webm?dl=0
> 
> * When moving focus with the keyboard, the mouse cursor stays still in Wayland (where in Xorg, it moves to the top-left corner of the selected window.)

* The GNOME Favourites (appplications on the left, after pressing Super- / Win- Key) get altered after installing pop-shell via `./rebuild.sh ` on Fedora 32. Prior I have had Firefox, Thunderbird, Files and Terminal, after the installation Terminal was missing which means my last Favourite got removed during install.

PR #519, opened 2020-08-28
feat: logging levels for development (Default OFF)

jmmaranan:
- Added logging levels to help control the noise in journal. And let any extension developers to toggle this on their own. I had to add the current value in the panel icon, to give a first glance.
- Some of the current log imports had been renamed to just `log`.

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -221,5 +221,17 @@
             <default>'rgba(251, 184, 108, 1)'</default>
             <summary>The current active-hint-color in RGBA</summary>
         </key>
+        <key type="u" name="log-level">
+            <default>0</default>
+            <summary>
+                Derive some log4j level/order
+                0 - OFF
+                1 - ERROR
+                2 - WARN
+                3 - INFO
+                4 - DEBUG
+            </summary>
+        </key>
+
     </schema>
 </schemalist>
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,7 @@ const Me = imports.misc.extensionUtils.getCurrentExtension();
 const { Gio, GLib } = imports.gi;
 
 import * as error from 'error';
-import * as Log from 'log';
+import * as log from 'log';
 import * as result from 'result';
 
 import type { Result } from 'result';
@@ -62,7 +62,7 @@ export class Config {
         const conf = Config.from_config();
 
         if (conf.kind === 2) {
-            Log.error(`failed to open pop-shell config: ${conf.value.format()}`);
+            log.error(`failed to open pop-shell config: ${conf.value.format()}`);
             return;
         }
 
@@ -78,7 +78,7 @@ export class Config {
         try {
             return JSON.parse(json);
         } catch (error) {
-            Log.error(`failed to parse config: ${error}`);
+            log.error(`failed to parse config: ${error}`);
             return new Config();
         }
     }
@@ -140,7 +140,7 @@ export class Config {
     sync_to_disk() {
         const result = Config.write(this.to_json());
         if (result.kind === 2) {
-            Log.error(`failed to sync disk to config: ${result.value.format()}`);
+            log.error(`failed to sync disk to config: ${result.value.format()}`);
         }
     }
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -8,7 +8,7 @@ import * as Focus from 'focus';
 import * as GrabOp from 'grab_op';
 import * as Keybindings from 'keybindings';
 import * as Lib from 'lib';
-import * as Log from 'log';
+import * as log from 'log';
 import * as PanelSettings from 'panel_settings';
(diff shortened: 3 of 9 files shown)


schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
```diff
@@ -174,5 +174,17 @@
             <default><![CDATA[['<Primary>Right','<Primary>l']]]></default>
             <summary>Swap window right</summary>
         </key>
+        <key type="u" name="log-level">
+            <default>3</default>
```
mmstick, line 178: I think we should default to OFF. People can choose what the log when they're actively debugging it.

jmmaranan: Defaulted it to OFF

mmstick (approved):

jacobgkau (approved):
Log levels seem to be working as expected. No other regressions.

I found it a little strange that we display the log level in the panel drop-down, but it's only adjustable in the extension settings. We may want to consider either having that option be adjustable in the drop-down, or having it open the extension settings when clicked.

mmstick:
We should remove it from the panel since this feature is mainly of use for development and debugging, rather than user-facing.

jmmaranan:
Removed the logging status on panel.

@mmstick - a bit off topic. Is there a plan somewhere to organize/expose the settings in prefs.js? I am seeing a trend in user request for customizability. I think that will solve us trying to re-balance some config versus user tweaking themselves and just provide a reset all? Need UX help on how would that be versus just appending all other settings as they come in prefs.js.

mmstick:
Everything in prefs.js is either experimental or would be clutter in the panel menu. The panel menu should have as few options, as needed by the user day to day.

jmmaranan:
I am not pertaining to cluttering the panel menu. More on organizing prefs.js for keyboard bindings for the shell, adjusting settings for delay, etc. Provide a quick jump to prefs.js on panel menu. Arc-menu does this very well.

mmstick:
The idea is that you can reach extension prefs with the native Extensions app in GNOME. So no real need to expose it in the panel. There's no need for keyboard bindings in prefs.js since this is made available in GNOME Settings > Keyboard and Pop Shell Shortcuts.

jmmaranan:
I will send something for the idea once have something working. Logging should be ready.

WatchMkr:
Besides access to certain features, the panels intent is to help users with learning a very new feature and way of computing. I imagine it not being necessary eventually... though that's on a very long timeline.

Panel and others tiling-related settings will reach GNOME Settings in Pop as part of broader desktop design work we're engaged in. We should continue to keep over-flow settings in the extension for other distros.

Where is documentation for prefs.js?

mmstick:
There's no documentation for what's in prefs.js

jmmaranan:
@WatchMkr - prefs.js is a free form GTK window. I've seen other extensions, organize what they want exposed beautifully.
I just thought it would lower the barrier of entry on using the extension. And just provide them a reset all if they went too far. I don't think all users are developers and can tinker with gsettings/dconf or what not.
This can be substituted with good documentation ( will take time for users ) or a good ui (best of both worlds).

WatchMkr:
@jmmaranan I'm looking forward to your ideas! I'd like to get a better understanding of what we're doing in prefs.js so we can work through the best places for those settings.

I don't think of Pop Shell as an extension but rather as an integrated part of the Pop!_OS whole. Many settings will integrate better when considering their functionality with the entire desktop in mind. Setting an application to always float might fit best in GNOME Settings > Applications for instance. For keyboard shortcuts, we just finished a major redesign of GNOME Settings > Keyboard which is in Pop now and landing in GNOME's next release. (For the most part... multiple keybindings for a single action is in Pop but won't make vanilla GNOME.)

jmmaranan:
> I'd like to get a better understanding of what we're doing in prefs.js so we can work through the best places for those settings.

- Pop-shell's keyboard shortcuts (unless can connect to gsettings of pop-shell to calculate conflicts)
- Configuration only for pop-shell, 
    - Tweener update delay (when you drag via mouse, the animation that swaps it back into the tiling group after releasing)
    - Active hint thickness?
- Re-organizing what is already there. I believe the list can grow.

> I don't think of Pop Shell as an extension but rather as an integrated part of the Pop!_OS whole. Many settings will integrate better when considering their functionality with the entire desktop in mind. Setting an application to always float might fit best in GNOME Settings > Applications for instance. For keyboard shortcuts, we just finished a major redesign of GNOME Settings > Keyboard which is in Pop now and landing in GNOME's next release.

I am seeing folks who wanted to modify the shortcuts only appearing on pop-shell but it is not visible on GCC. From a technical standpoint, It might look like we just needed to connect to the pop-shell gsettings schema from GCC, however I think this will break anyone using the GCC that does not have pop-shell if it goes upstream. And at the same time folks who are distributing pop-shell as an extension on other distros.

The prefs.js looks like the common interface for settings even though it can be done in GCC. And it would be portable if in prefs.js as it was I guess intended?

PR #523, opened 2020-08-29
fix: handle null entity on reparent, check parent before re-stacking siblings

jmmaranan:
Some small cleanup when moving stacked windows on different workspaces.


mmstick:
Thanks!

PR #528, opened 2020-09-04
fix: check primary monitor workspace, no reflow

### LINKED ISSUE #246: Unable to move a window to a second monitor that has a floating window on it
```
NAME="Pop!_OS"
VERSION="20.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
VERSION_ID="20.04"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
LOGO=distributor-logo-pop-os
``` 
 
```
pop-shell:
  Installed: 0.1.0~1588274983~20.04~586efcc
  Candidate: 0.1.0~1588274983~20.04~586efcc
  Version table:
 *** 0.1.0~1588274983~20.04~586efcc 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main i386 Packages
        100 /var/lib/dpkg/status
```

**Issue/Bug Description:**
Moving windows to other screens is inconsistent when the second screen has a floating window on it. Moving with either adjument mode, or just using Super+Shift+Left/Right doesn't do anything.

Dual monitor setup, open a window on one monitor and set it to floating, and try moving another window from the other monitor with keybindings.

**Expected behavior:**
Windows should move seamlessly between monitors even if the second monitor contains any active floating windows.

**Other Notes:**
Moving with the mouse works as expected, and it seems like the only workaround as of right now.

### LINKED ISSUE #522: "move window one monitor to the right/left" doesn't work with tiling enabled

jmmaranan:
Do not reflow when workspace primary is disabled. 

Hopefully: 
- Fixes #522
- Fixes #246 - looks like a dupe of 522 or vice versa

I do not have 3 monitors.
Need testing on primary, secondary, secondary or secondary, primary, secondary setups.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -251,7 +251,11 @@ export class Ext extends Ecs.System<ExtEvent> {
 
                     case WindowEvent.Size:
                         if (this.auto_tiler && !win.is_maximized() && !win.meta.is_fullscreen()) {
-                            this.auto_tiler.reflow(this, win.entity);
+                            if (!this.settings.workspaces_only_on_primary()) {
+                                this.on_workspace_changed(win);
+                            } else {
+                                this.auto_tiler.reflow(this, win.entity);
+                            }
                         }
                         break
 


mmstick (approved):

jacobgkau:
The keyboard shortcuts to move windows between displays still don't work as expected with tiling enabled after this change.

With two displays of different sizes, and the primary display being the smaller one, I can move a window from primary to secondary using Super-Shift-Right, but not back the other way with Super-Shift-Left: https://www.dropbox.com/s/gvyqzqbf21yw4jj/shell-pr528-screencast2.webm?dl=0 

As you can see, when I try to move from the larger secondary display to the smaller primary display, the window moves a few pixels. Once that happens, it doesn't move any farther when I try pressing Super-Shift-Right/Left again; I need to enter adjustment mode to get it un-stuck.

Here's a three-display example where the primary display (center) is one of the larger displays: https://www.dropbox.com/s/20dw3pdpk0cqr65/shell-pr528-screencast6.webm?dl=0 

- The window gets stuck when moving from a secondary display to a primary display of the same size (which has less usable space due to the top bar.)
- When I enter adjustment mode and press the left arrow key once, the window moves all the way to the left-most display, which suggests that Shell thought the window was on the center display at that point (even though it looked like it was still on the right-most display.)

In this layout, I'm able to move between the three displays successfully because the center display is tiling the windows small enough to fit on either of the side displays: https://www.dropbox.com/s/gcsa14tho7ydn1t/shell-pr528-screencast5.webm?dl=0 

Since I can see in that last video that the window's being moved to a smaller display before it's being resized, that makes me think the top bar is the only thing preventing the window from being moved in the other cases.

jmmaranan:
Thanks @jacobgkau - the top bar or panel on the primary is what seems to be blocking the windows from moving. If the panel is displaying across all windows (Dash-To-Panel), it works on two monitors. 

Would you be able to try the test cases in a 3 monitor setup with any extension that spans the Panel across monitors? I will try also this in a built-in display and 2 monitors. Waiting for the USB-to-USB-C adapter.

Another thing is that this looks like a standard Mutter or Gnome-Shell, `move windows to monitor` functionality based on the keybindings, so this might be a bug upstream.

jacobgkau:
> Would you be able to try the test cases in a 3 monitor setup with any extension that spans the Panel across monitors?

With three monitors and Dash to Panel, the issue occurs when going from a larger display to a smaller display. The window does shift up (away from the panel) instead of down: https://www.dropbox.com/s/h2lw6snwsyqohh3/shell-pr528-screencast7.webm?dl=0 
 
> Another thing is that this looks like a standard Mutter or Gnome-Shell, `move windows to monitor` functionality based on the keybindings, so this might be a bug upstream.

I think you're correct. After disabling the Pop Shel extension and restarting the shell, I'm able to reproduce the issue. Large and small display: https://www.dropbox.com/s/cwwze9t7x3qbqvr/shell-pr528-screencast8.webm?dl=0 

Two large displays: https://www.dropbox.com/s/ufw9x8gtkghfelf/shell-pr528-screencast9.webm?dl=0 

As you can see, the issue doesn't occur until the window is nearly filling up the source display. The issue also doesn't occur when a window is maximized. It seems Pop Shell is exposing this bug more prominently because it makes it very easy/common to have a window take up almost an entire display (with less than a panel's worth of gap), but not be maximized.

I see the following upstream issues open for this:

https://gitlab.gnome.org/GNOME/mutter/-/issues/455 

https://gitlab.gnome.org/GNOME/mutter/-/issues/477 

As a side note, it doesn't seem entirely incorrect for GNOME to refuse to move a window to a display where it's not going to fit, although that's inconsistent with being able to move a window to a smaller display when there's not a panel in the way.

mmstick:
It's definitely a GNOME Shell issue, because `<Super>Shift`-based movements are native GNOME Shell keyboard shortcuts, and we don't have much control over the code they execute for these actions.

jmmaranan:
@jacobgkau - your observations seem to be correct. Without the tiling, the window is moved to the same allocation/coordinates as of the previous monitor and scaled by the Window Manager if they are in different resolutions.

@mmstick - I think binding the keyboard to Pop Shell's window management left and right movement will fix this in the time being if we would want the functionality to be seamless. Or a vanilla implementation of the windowmanagement left/right. Thoughts?

mmstick:
Having our own implementation of moving windows across monitors would be the way to go, and doing away with the bindings to GNOME's. Only way to be sure we have control over how windows move around workspaces and displays.

jmmaranan:
@mmstick - then the current change only fixes moving out to a monitor that it thinks it can accommodate the current window allocation. Putting this in DRAFT or WIP.

mmstick:
Moving windows between monitors should be fixed now. Windows are shrunk to find on the monitor they're moved to if they happen to be too large for the monitor.

@jmmaranan Could you rebase these changes onto stacking_focal?

jmmaranan:
@mmstick - just to be sure, had put this on #543

PR #530, opened 2020-09-04
Update rebuild.sh

samuela:
So that there's a little space between "(y/n)" and the input prompt.


mmstick:
Could you also fix the conditionals so that they aren't using `[[`? Since this is a feature of bash, and is not POSIX-compliant.

samuela:
I'm not very familiar with this shell scripting stuff, what's the alternative to `[[ ... ]]`?

mmstick:
`[[` is essentially equivalent to `test`

samuela:
@mmstick Ok, done in https://github.com/pop-os/shell/pull/530/commits/8d930372ff1b4e02b980c16f9608e45b7c219c3b.

Issue #513, opened 2020-08-23
GUI for adding windows to the floating list


domwoolf:
Pop shell does not always correctly identify whether a window should tile.  E.g. pop-ups should typically not tile. If the program that generates a pop-up correctly identifies it as such, then Pop-Shell (correctly) does not tile them.  However, some software generates pop-ups that are not identified as such.  It is extremely useful for the user to be able to specify their desired behaviour for these windows.

Looking through the source code I discovered that Pop-shell does read window-specific rules from a file config.json.  Using this I was able to generate by trial and error a json file that sets the "window_shall_float" flag to false for a specified window title.

However, this ability to blacklist windows from tiling is not documented.  A useful improvement to Pop-shell would be to either document clearly how users can edit the config file for themselves, or (better) provide a UI interface to choose whether to blacklist a window by adding it to this config. 

My config.json file now looks like this:

```
{
  "float": [
    {
      "class": "Zotero",
      "title": "Quick Format Citation"
    }
  ],
  "log_on_focus": false
}

```

mmstick:
If the popup window describes itself as a modal window with a WM_HINT, then it will automatically be ignored for tiling. You'll notice that native GTK and Qt applications typically observe this rule and tile correctly, but applications using Electron do not. Zotero is an Electron application, and therefore it does not set the correct WM_HINT values. So the only way to handle them is to have a list of known windows that should float.

mmstick:
Also, `window_shall_float` is not a variable. It's a method. The only two variables in the config are `log_on_focus` and `float` right now. The class and title are parsed as regular expressions.

domwoolf:
Thanks @mmstick for the feedback.  Until such time as this enhancement gets applied, is there any chance you could provide  an example here of how the config file should look in order to blacklist a window of class= "xxxx" and title = "yyyy"?  That would really help document the correct method for myself and others trying to implement it now.

mmstick:
The values you set for `float` are correct. Each rule object contains two optional keys: `class` and `title`. Setting `log_on_focus` to `true` will print the class and title of a window whenever a window is focused.

mmstick:
You may also submit a PR to add these values our list of defaults

jmbuhr:
@domwoolf , where does this config.json file live? :)

domwoolf:
@jmbuhr In my installation the config.json is in ~/.config/pop-shell

jmmaranan:
[...]

PR #534, opened 2020-09-05
Don't tile enpass assistant

Chrysostomus:
Don't tile enpass assistant, it is a popup window even though it does not tell it.


Chrysostomus:
https://github.com/pop-os/shell/issues/513#issuecomment-678826704

Pull request submitted as suggested here. Enpass assistant has a small maximum size and doesn't fill the space that it is tiled to. It is meant to be a pop-up dialog and autocloses if it loses focus, so having it float is kinda good for the usability.

PR #537, opened 2020-09-07
add shebang to rebuild.sh

Daniel-E-B:


mmstick:
I'd honestly rather the script be POSIX instead of Bash

PR #542, opened 2020-09-12
fix: Watch scale-factor changes and update gaps on changes

### LINKED ISSUE #492: Display scaling doesn't change gap width without gnome-shell reset
(pretty minor bug)

 Pop!_OS 20.04
 
 
Pop-shell version: `0.1.0~1597190322~20.04~149061a`

**Issue/Bug Description:**
Setting a different scaling factor does not scale gap sizes until I manually restart gnome-shell.

Starting with tiling on and the default gap size of 2:
Screenshot from 2020-08-12 10-16-35

Change scaling factor to 150% in Display settings and the gap sizes will look too small:
Screenshot from 2020-08-12 10-16-47

After restarting gnome-shell (alt+F2, r, enter), the gaps look the way I would expect:
Screenshot from 2020-08-12 10-16-55

Setting it back to 100% does the same thing, but the larger gap size from 150% scaling persists:
Screenshot from 2020-08-12 10-17-17

and another gnome-shell reset brings it back down to what I would expect: 
Screenshot from 2020-08-12 10-17-23

**Expected behavior:**
Gap size should change with scaling without requiring a gnome-shell restart.

mmstick:
Fixes #492

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -592,18 +592,21 @@ export class Ext extends Ecs.System<ExtEvent> {
         let prev_gap = this.gap_inner_prev / 4 / this.dpi;
 
         if (current != prev_gap) {
-            if (this.auto_tiler) {
-                for (const [entity,] of this.auto_tiler.forest.toplevel.values()) {
-                    const fork = this.auto_tiler.forest.forks.get(entity);
-                    if (fork) {
-                        this.auto_tiler.tile(this, fork, fork.area);
-                    }
-                }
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Confirmed this fixes https://github.com/pop-os/shell/issues/492, no regressions. [shell-pr542.txt](https://github.com/pop-os/shell/files/5220605/shell-pr542.txt)

PR #544, opened 2020-09-13
active-hint enhancements

### LINKED ISSUE #219: "Show active hint" almost entirely disappears when the gap is set to "0".
During the usability testing three users have pointed out that setting the gap to 0 would almost entirely hide it from view. Those users preferred to have as much screen space as possible without any gaps.

One of the possible solutions is to disable the hint when someone sets the gap to "0". But that might result in the confusion and frustration if a user toggled the hint on before setting the gaps to "0". We should probably avoid a user thinking it is some sort of bug. 

I think it might also be a question of adjusting the hint overall. Some participants in the testing and, to be honest, me too, found the hint to bee too distracting. It might be the question of color but it might also be the question of the thickness of the hint. I will look into it and see if I can come up with any other ideas.

jmmaranan:
Fixes #219, 

0 - gapped:
image

maximized:
image

### DIFF
--- a/dark.css
+++ b/dark.css
@@ -1,9 +1,16 @@
 .pop-shell-active-hint {
-    border-width: 5px;
     border-style: solid;
     border-color: #FBB86C;
     border-radius: 5px;
     box-shadow: inset 0 0 0 1px rgba(24, 23, 23, 0)
+}
+
+.pop-shell-border-normal {
+    border-width: 3px;
+}
+
+.pop-shell-border-maximize {
+    border-width: 3px;
 }
 
 .pop-shell-search-element:select{
--- a/light.css
+++ b/light.css
@@ -1,9 +1,16 @@
 .pop-shell-active-hint {
-    border-width: 5px;
     border-style: solid;
     border-color: #FFAD00;
     border-radius: 5px;
     box-shadow: inset 0 0 0 1px rgba(200, 200, 200, 0);
+}
+
+.pop-shell-border-normal {
+    border-width: 3px;
+}
+
+.pop-shell-border-maximize {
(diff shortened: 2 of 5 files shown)


mmstick:
What's left to get this ready? Would like to merge before stacking

jmmaranan:
The smart gap is not being identified properly. I added a function in window.ts and it doesn't work well.

jmmaranan:
@mmstick - since #219 did not originally define smart gaps, would it be acceptable? And also smart gaps is an experimental feature.

jmmaranan:
I will cleanup the logging later this afternoon on that last commit.

mmstick:
If necessary, I can add what's needed to get smart gaps working with this

jmmaranan:
Sure. I will gladly take your help on that. I have a fix on the window when detached the other windows follow it. I was hoping to put that in another MR

jmmaranan:
But I cannot find the cause for why it doesn't detect the smart gapped flag

jacobgkau (approved):
Looks good. I will change the verbiage in the testing document to reflect that the active hint should no longer disappear for maximized windows in https://github.com/pop-os/shell/pull/526.

Full-screened apps still hide their hint, which is correct behavior.

I wanted to mention that when maximizing/un-maximizing (and also when using Ctrl-Super-Left/Right), the active hint moves to its new position before the window gets there (the active hint moves instantly, while the window's movement is animated by GNOME Shell.) This is the same behavior as current stable, it's just seen in more places because the hint was hidden on maximize/half-snapped before and now it's not. I don't think this is any more visually jarring than the rest of pop-shell's movements at this time.

PR #554, opened 2020-09-20
feat: active hint color selector

### LINKED ISSUE #462: Active Hint Color Selector
As a feature improvement to enhance the aesthetic, a selector to choose the color of the active hint. The choices could include the current Pop-orange, Pop-blue and a light grey that suit the Dark theme. These three colors should be adequate to make most theme choices a bit easier on the eye. The Active Hint Color Selector could live in the Pop-Shell settings found in the Extensions App.

jmmaranan:
Fixes #462 -
https://www.dropbox.com/s/yrom2fwq64c4d42/Peek%202020-09-20%2018-10.webm?dl=0


Makefile
```diff
@@ -81,6 +81,7 @@ schemas: schemas/gschemas.compiled
 	touch $@
 
 schemas/gschemas.compiled: schemas/*.gschema.xml
+	sudo cp schemas/*.xml /usr/share/glib-2.0/schemas && sudo glib-compile-schemas /usr/share/glib-2.0/schemas
```
mmstick, line 84: This will cause debian packaging to fail. You're not permitted to move files into system paths.

jmmaranan: Sure - I will check why the Gio.Settings API does not work but the command line does recognize a user path.
https://github.com/pop-os/shell/pull/554/files/117f4bc3e1e226ba35aa353852e1a4b83b075837#diff-e4512274660514546724494dd7ae0a43R36

jmmaranan: Looks like an oversight, I have to use GSettingsSchemaSource to build the object itself so it has the standard GSettings.Backend.

mmstick (dismissed):

jacobgkau (dismissed):
Looks good. [shell-pr554.txt](https://github.com/pop-os/shell/files/5258338/shell-pr554.txt)

It looks like either an rgb(w,x,y) or rgba(w,x,y,z) value works in gsettings, although the alpha channel in an rgba value is ignored (1 or above is opaque, 0 is entirely transparent), as has been noted in the code comments.

If I set an invalid value in the gsetting, the little square that's supposed to show the current color disappears:

shell-pr554-screenshot1

I can still click in the empty space to set a valid color again, it's just difficult to find when invisible. The active hint itself goes back to the default orange in this case. I wouldn't imagine a user accidentally setting this option via the gsettings command, and this would never happen if they're using the color palette we provide (which allows setting any color via the hex code-- there's really no reason to use the gsettings command), so I don't think this is critical.

jmmaranan:
Thanks. I can add the css value on file if the gsettings is invalid. If it becomes problematic.

jmmaranan:
The CSS alpha is actually being applied using the bottom slider of the color dialog. Cleaned up a bit based on the coding style for the settings, so the setter/getter for the color is centralized just like the other settings.js.

jacobgkau (approved):

PR #555, opened 2020-09-21
fix: tweener stutter

### LINKED ISSUE #547: Double update / delay
Hi all, 
I've noticed that when a windows gets moved around or the order gets switched, or more generally whenever the layout of multiple windows is affected, that the repositioning and resizing of the non-focused windows happens after the repositioning and resizing of the focused window on which the action was performed. Basically it causes a bit of a stutter is what it feels like. Although it's not a huge issue for me, it does make the whole tiling system feel a lot less smooth and kind of distracting. I've turned off animations in gnome tweaks hoping that would solve it, but all it does is make the double update finish a little faster. 

Would it be possible to get layout updates done in one single update?

jmmaranan:
closes #547 - remove the extra animation after done tweening.
Also, need to check scale-y correctly when swapping vertically.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1138,8 +1138,6 @@ export class Ext extends Ecs.System<ExtEvent> {
         let win = this.get_window(window);
         if (win) {
             const entity = win.entity;
-            if (actor)
-                actor.hide();
             if (win && win.border)
                 win.border.hide();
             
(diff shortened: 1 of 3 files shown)


mmstick (approved):

jmmaranan:
This needs prefs.js input for changing the delay - subjective/preference to some users.

jmmaranan:
Closing this for now. Planning to work on keybinding management inside pop-shell's prefs.js

PR #568, opened 2020-09-25
fix: gnome-terminal -e option being deprecated

### LINKED ISSUE #560: -e option is depreciated for gnome-terminal
in src/launcherext.ts -e option is used after opening gnome-terminal this option is depreciated. The new syntax should be: 
```spawnCommandLine(`${terminal} -- sh -c '${cmd}; echo "Press to exit"; read t'`);```

This would only work for gnome-terminal, for xterm -e option is still supported

mmstick:
Closes #560

### DIFF
--- a/src/launcherext.ts
+++ b/src/launcherext.ts
@@ -14,7 +14,7 @@ import type { Search } from 'search';
 
 const DEFAULT_ICON_SIZE = 34;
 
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Behavior and logging seem identical before and after the change. Regression testing passed.

PR #569, opened 2020-09-26
improv(rebuild.sh): offer to install TypeScript with npm if missing or too old

toreanderson:
This should make it easier for users of distributions with too old TypeScript
versions (e.g., Fedora 32) to try Pop Shell.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ debian/*
 !debian/rules
 .confirm_shortcut_change
 .vscode
+node_modules
+package-lock.json
--- a/rebuild.sh
+++ b/rebuild.sh
@@ -12,6 +12,20 @@ else
     echo "Shortcut change already confirmed"
 fi
 
+# Ensure TypeScript is new enough, offer to install with npm if not
+export PATH="./node_modules/.bin:$PATH"
+REQ_TSC_VER="3.8"
+CUR_TSC_VER=$(tsc --version 2>/dev/null || echo 0)
+CUR_TSC_VER=${CUR_TSC_VER##* }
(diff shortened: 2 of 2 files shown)


mmstick:
Had to revert this because it breaks the script

jmmaranan:
It does break on sh and zsh. And it always asks to install tsc when I am already on v4.x

toreanderson:
Interesting, I will try to figure out how it breaks for you when using `sh` and re-post. (It worked for me on my Fedora 32, for what it is worth.)

I am assuming compatibility with other shells (such as `zsh`) is not a goal, given the `.sh` file name suffix.

jmmaranan:
I think some of the operators used are bash specific? Don't mind zsh. I can invoke bash explicitly in front of the script.

@mmstick what's the shell default for the build?

mmstick:
`sh` is a strictly POSIX-compliant shell. Dash in the Debian/Ubuntu case. So only POSIX syntax is supported, which means no Bash extensions.

toreanderson:
Thanks, I think it is the `<<<` operator it doesn't like:

```
[:~] 2 $ podman run -it --rm debian
root@fecc3388b76f:/# echo $0
bash
root@fecc3388b76f:/# cat <<< input
input
root@fecc3388b76f:/# sh
# echo $0
sh
# cat <<< input
sh: 2: Syntax error: redirection unexpected
```

I'll update the PR shortly.

That said, any particular reason not to simply use `#!/bin/bash` in the top of the script, or is it a realistic scenario that someone would want to install Pop Shell on a system without Bash?

PR #579, opened 2020-09-29
feat: Restore window dimensions on detach

### LINKED ISSUE #379: Tiling overwrites default application sizes
**Distribution:**
NAME="Pop!_OS"
VERSION="20.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
VERSION_ID="20.04"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
LOGO=distributor-logo-pop-os
 
 
**Related Application and/or Package Version**:
  Installed: 0.1.0~1589754807~20.04~e5f3b4d
  Candidate: 0.1.0~1589754807~20.04~e5f3b4d
  Version table:
     0.1.0~1589754807~20.04~e5f3b4d 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu focal/main i386 Packages
        100 /var/lib/dpkg/status

**Issue/Bug Description:**
After using the tiling mode and switching it off, the next time you launch an application, it opens with the size and in the position that is was, when tiling was enabled. Which can be annoying, because without the other applications being open it does not form a nice grid, and looks a bit funny.

**Steps to reproduce :**
1. Have an application sized the way you like it.
2. Turn on tiling.
3. Close the application.
4. Turn tiling off.
5. Reopen the application.
6. The application opens with its previous "tiling size' instead of the one you liked.

**Expected behavior:**
In my opinion the size and position of an application should be ignored, when tiling was enabled and applications should open with the size and position what they had the last time before tiling was enabled. It would make the experience much better.

**Other Notes:**
Tiling is awesome, I love pop-shell very much, thanks for making my life easier!

mmstick:
Closes #379

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -209,8 +209,14 @@ export class AutoTiler {
     /** Destroy all widgets owned by this object. Call before dropping. */
     destroy(ext: Ext) {
         for (const [, [fent,]] of this.forest.toplevel) {
-            for (const window of this.forest.iter(fent, NodeKind.WINDOW)) {
-                this.forest.on_detach((window.inner as node.NodeWindow).entity)
+            for (const node of this.forest.iter(fent)) {
+                if (node.inner.kind === 2) {
+                    this.forest.on_detach(node.inner.entity)
+                } else if (node.inner.kind === 3) {
+                    for (const window of node.inner.entities) {
+                        this.forest.on_detach(window)
+                    }
+                }
             }
         }
 
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -75,10 +75,10 @@ export class AutoTiler {
         }
 
         const a_fn = a_fork.replace_window(ext, a_win, b_win);
-        this.attached.insert(b, a_ent);
+        this.forest.on_attach(a_ent, b);
 
         const b_fn = b_fork.replace_window(ext, b_win, a_win);
-        this.attached.insert(a, b_ent);
+        this.forest.on_attach(b_ent, a);
 
         if (a_fn) a_fn();
         if (b_fn) b_fn();
@@ -138,7 +138,7 @@ export class AutoTiler {
         }
 
         const [entity, fork] = this.forest.create_toplevel(win.entity, rect.clone(), workspace_id)
-        this.attached.insert(win.entity, entity);
+        this.forest.on_attach(entity, win.entity);
         fork.smart_gapped = smart_gaps;
         win.smart_gapped = smart_gaps;
 
@@ -208,6 +208,12 @@ export class AutoTiler {
 
     /** Destroy all widgets owned by this object. Call before dropping. */
     destroy(ext: Ext) {
+        for (const [, [fent,]] of this.forest.toplevel) {
+            for (const window of this.forest.iter(fent, NodeKind.WINDOW)) {
+                this.forest.on_detach((window.inner as node.NodeWindow).entity)
+            }
+        }
+
         for (const stack of this.forest.stacks.values()) stack.destroy();
 
         for (const window of ext.windows.values()) {
@@ -233,6 +239,8 @@ export class AutoTiler {
 
                 this.tile(ext, fork, fork.area);
             }
+
+            ext.windows.with(win, (info) => info.ignore_detach = false);
         });
     }
 
--- a/src/ecs.ts
+++ b/src/ecs.ts
@@ -106,6 +106,12 @@ export class Storage<T> {
         this.store[id] = [gen, component];
     }
 
+    /** Check if the storage is empty */
+    is_empty(): boolean {
+        for (const slot of this.store) if (slot) return false;
+        return true;
+    }
+
     /// Removes the component for this entity, if it exists
     remove(entity: Entity): T | null {
         const comp = this.get(entity);
--- a/src/events.ts
+++ b/src/events.ts
@@ -67,8 +67,7 @@ export function global(event: GlobalEvent): GlobalEventTag {
 }
 
 export function window_move(ext: Ext, window: Window.ShellWindow, rect: Rectangular): ManagedWindow {
-    if (!window.movement) ext.windows_moving += 1;
-    window.movement = rect;
+    ext.movements.insert(window.entity, rect);
     return { tag: 2, window, kind: { tag: 1 } };
 }
 
--- a/src/events.ts
+++ b/src/events.ts
@@ -3,6 +3,8 @@ const Me = imports.misc.extensionUtils.getCurrentExtension();
 
 import * as Window from 'window';
 
+import type { Ext } from 'extension';
+
 /** Type representing all possible events handled by the extension's system. */
 export type ExtEvent = GenericCallback
     | ManagedWindow
@@ -44,7 +46,6 @@ export enum GlobalEvent {
 
 export interface Movement {
     tag: 1;
-    rect: Rectangular;
 }
 
 export interface Basic {
@@ -65,8 +66,10 @@ export function global(event: GlobalEvent): GlobalEventTag {
     return { tag: 4, event };
 }
 
-export function window_move(window: Window.ShellWindow, rect: Rectangular): ManagedWindow {
-    return { tag: 2, window, kind: { tag: 1, rect } };
+export function window_move(ext: Ext, window: Window.ShellWindow, rect: Rectangular): ManagedWindow {
+    if (!window.movement) ext.windows_moving += 1;
+    window.movement = rect;
+    return { tag: 2, window, kind: { tag: 1 } };
 }
 
 /** Utility function for creating the an ExtEvent */
--- a/src/events.ts
+++ b/src/events.ts
@@ -2,7 +2,6 @@
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 
 import * as Window from 'window';
-import { Rectangle } from './rectangle';
 
 /** Type representing all possible events handled by the extension's system. */
 export type ExtEvent = GenericCallback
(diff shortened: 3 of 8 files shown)


jacobgkau (commented):
This doesn't seem to work entirely reliably.

- Nautilus still stays at its tiled position when resized/restarted after resizing: https://www.dropbox.com/s/3vu7ugtmweuuav2/shell-pr579-screencast1.webm?dl=0 
- If a window is quit while tiling is on, it opens in its tiled size when launched while tiling is off (this example uses Audacity) -- this sounded like what #379 was referring to: https://www.dropbox.com/s/8c770gkoenlhf5u/shell-pr579-screencast2.webm?dl=0 
- Windows seem to forget their previous dimensions if they're moved between workspaces while tiling is enabled: https://www.dropbox.com/s/b76u6bpxpmij9dq/shell-pr579-screencast3.webm?dl=0 

While testing with Firefox and Audacity, I've also had multiple cases where only one window would remember its position after enabling/disabling tiling (sometimes one, sometimes the other.) Examples:

- Audacity misbehaving: https://www.dropbox.com/s/kqeqe6hj9ylkaus/shell-pr579-screencast4.webm?dl=0 
- Firefox misbehaving: https://www.dropbox.com/s/dd29qat5txcm7u9/shell-pr579-screencast5.webm?dl=0 

If any of this is expected, I would need more detailed information about the expected behavior.

mmstick:
I'm looking into it. This is helpful

mmstick:
I was able to fix everything but tiles that are destroyed while being auto-tiled. By the time we get the destroy signal from a `Meta.Window`, the window has already been destroyed, and it's no longer possible to resize it so that GNOME remembers its original location.

jacobgkau:
Understood re: destroyed while auto-tiled. May want to make a note about that in the issue so it doesn't get re-opened, since the test case described will still "fail."

Moving between workspaces still loses the saved size: https://www.dropbox.com/s/16w2ageovlutmof/shell-pr579-screencast6.webm?dl=0 Is this one something that can be worked around, or will it also not work?

jacobgkau:
Positions are now correctly being remembered even when windows are moved between workspaces. However, I'm now seeing a ghost window issue: https://www.dropbox.com/s/qweccsfbp3zyx50/shell-pr579-screencast7.webm?dl=0 

This is on `ffac96f`, and is near the beginning of the testing.md procedure. After arranging the windows and then swapping two of the windows, I'm not able to resize certain windows (in this example, the stack won't resize, and neither will the terminal.) When I try to move one of the windows that won't resize, I get a ghost window. This is not happening in master.

Here's an example without a stack: https://www.dropbox.com/s/p5bmh7l9s50boel/shell-pr579-screencast8.webm?dl=0 It looks like the swap is what causes the issue, as I'm able to resize windows fine until I swap two windows (happens with both differently-sized and same-sized windows.)

mmstick:
@jacobgkau I think I've fixed the problem. I'm able to swap and resize without issues now.

jacobgkau:
Enabling and disabling tiling in quick succession breaks the pre-tiling memory: https://www.dropbox.com/s/fryjv5eeqyfdim5/shell-pr579-screencast9.webm?dl=0 Is this an expected performance limitation?

mmstick:
I'll see if I can work around that scenario

mmstick:
@jacobgkau That last commit will have the shell track when windows are moving, and delay toggling of tiling when they are moving.

jacobgkau:
That fix did work to prevent the shell from forgetting sizes when quickly toggling on and back off. However, I'm now not able to turn tiling off after closing a window with `Super+Q`. (Neither the keyboard shortcut nor the on/off switch in the top-right menu work; clicking the on/off switch flips the switch, but doesn't turn off tiling. The keyboard shortcut does not flip the switch.)

mmstick:
@jacobgkau Got that one fixed too

jacobgkau:
Regression testing is passing again. I'm seeing some interesting behavior with remembered positions for stacked windows: https://www.dropbox.com/s/shvzxpzvvy56gzs/shell-pr579-screencast10.webm?dl=0 

It seems the window's not properly restored to its prior position if I disable tiling while it was stacked; however, if I re-enable and re-disable tiling, it goes back to its original position. This doesn't seem intentional.

mmstick:
Toggling tiling wasn't triggering the `on_detach()` signals for stacked windows. Now working

jacobgkau:
@mmstick On `d60646a`, stacked windows restore their positions, but non-stacked windows do not restore their positions.

mmstick:
Fixed

jacobgkau:
I am getting misplaced windows if I press the arrow key too quickly for the shell to keep up: https://www.dropbox.com/s/w9pv6d3ggqs0l3s/shell-pr579-screencast11.webm?dl=0 

In comparison, master (`50d953a`) simply swaps the windows if I try to move too fast, I don't end up with any empty spaces or overlapping windows: https://www.dropbox.com/s/ctpl3ks7x8t85hy/shell-pr579-screencast12.webm?dl=0

jackpot51 (approved):

mmstick:
Took a while, but I've rearranged the way window movements are scheduled, to eliminate any risk of operations being applied out-of-order due to window animations, and ensuring a delay between movements of the same window to stop the animation quirks. This may fix the blurry windows too.

jacobgkau:
That did fix the visual glitch. Certain actions (especially repetitive ones) are noticeably slow now. For example, with tiling disabled, enter window management mode with `Super+Enter` and try to move a window around-- if I press the arrow keys fast enough (or even hold them down), then the window continues moving after I let go of the key. I can even exit window management mode, then re-enter window management mode, and the window starts moving again without having to touch the arrow keys. I don't think it's okay for actions to queue up like this: https://www.dropbox.com/s/85m5n1kuumxduu4/shell-pr579-video1.webm?dl=0 

Notice how the second time I entered window management mode, I didn't press the right arrow key at all, that movement was entirely left over from the first move. Even if movements are slow, I would expect them to at least be cleared/dropped when window management mode is exited.

I can also feel a difference moving windows around while tiling is enabled (it seems to be slower even when animations are disabled), but the above example is the easiest to quantify.

jacobgkau:
Thinking further, I realize that you made movements queue to solve a problem, and now I've called the queueing a problem. Maybe, while it makes sense to queue some actions (like moving a window in tiling mode), it doesn't make sense to queue others (like moving a window in non-tiling/floating mode), especially when the added delay hinders those features more than the queuing is helping them (moving in non-tiling/floating mode is basically instant on master.)

mmstick:
Non-tiling/floating windows aren't affected by this. The queue is only for when a window is moved by the auto-tiler, and when a movement is requested by the user through the tiler. I could limit the number of movements that will be scheduled, and drop the queue when exiting management mode. I can also increase the animation speed since the delay is based solely around that.

mmstick:
@jacobgkau See if these changes make the experience better

jacobgkau:
> Non-tiling/floating windows aren't affected by this. The queue is only for when a window is moved by the auto-tiler, and when a movement is requested by the user through the tiler.

I may be misunderstanding you, but moving windows using window management mode with auto-tiling disabled was most certainly affected by this. The issue I showed in my video above is not reproduceable on master/stable, and is also not reproduceable after the last three commits. Would this be included in "movement being requested by the user through the tiler?" 

The movement is at a more acceptable level now, although it's still slower than before. Here are two examples of what happens when I hold down the right arrow key before and after this branch:

- Moving a window across the screen on current stable: https://www.dropbox.com/s/3qkfdrn2ktk11vg/shell-pr579-screencast14.webm?dl=0 
- Moving a window across the screen on this branch: https://www.dropbox.com/s/e05a651xtf8a010/shell-pr579-screencast15.webm?dl=0 

The difference in speed is clearly visible. While that movement being slower isn't a big problem, the visual glitch from before last night's changes has reappeared: https://www.dropbox.com/s/as6ysk8jis4vd5v/shell-pr579-screencast16.webm?dl=0

There are two different things happening here:
1. If I try to move the window too quickly, it moves back to its previous location. This also happens on master.
2. When this occurs, the other window turns invisible until it's focused again. This does not happen on master, that's the regression.

I hope this clarifies things, let me know if there are limitations/expectations I'm not taking into account.

mmstick:
I can't seem to replicate a disappearing window, even with the same applications and layout.

jacobgkau:
Hmm. After switching to a separate machine (the addw2 with a relatively fresh install), I was able to make it happen on my second try after adding this branch and restarting the shell (it didn't happen on stable.)

I saw some cases earlier where the window didn't disappear until I exited window management mode: https://www.dropbox.com/s/5ax9jzl76tun62q/shell-pr579-screencast18.webm?dl=0 

I tried to get logs, but even set to `DEBUG`, I'm just seeing a single `gnome-shell[1965]: Window manager warning: Overwriting existing binding of keysym 6f with keysym 6f (keycode 20).` every time I exit window management mode, no other logging. Is there any other info I could get to help?

mmstick:
I'm going to try something out. A shot in the dark, but I have some suspicions about the cause.

mmstick:
The overlay movements should be back to normal speed. Hopefully this change will fix the disappearing windows too

jacobgkau (approved):
The visual glitch with disappearing windows (and the window swapping when moving too quickly glitch) is gone, so that's great. Moving windows around in tiling mode feels much snappier now. Moving floated/non-tiled windows in management mode is back to the way it was before.

Ran through regression testing again, and everything's looking good now. [shell-pr579-run6.txt](https://github.com/pop-os/shell/files/5337089/shell-pr579-run6.txt)

PR #580, opened 2020-09-29
feat: Floating window config dialog

### LINKED ISSUE #513: GUI for adding windows to the floating list

mmstick:
Closes #513

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -15,10 +15,10 @@ const { Err, Ok } = result;
 const CONF_DIR: string = GLib.get_home_dir() + "/.config/pop-shell"
 export var CONF_FILE: string = CONF_DIR + "/config.json"
 
-const DEFAULT_RULES: Array<FloatRule> =[
+export const DEFAULT_RULES: Array<FloatRule> = [
     { class: "Authy Desktop", },
-    { class: "Enpass", title: "Enpass Assistant"},
-    { class: "Zotero", title: "Quick Format Citation"},
+    { class: "Enpass", title: "Enpass Assistant" },
+    { class: "Zotero", title: "Quick Format Citation" },
     { class: "Com.github.donadigo.eddy", },
     { class: "Conky", },
     { class: "Gnome-screenshot", },
@@ -40,8 +40,12 @@ export class Config {
     /** Logs window details on focus of window */
     log_on_focus: boolean = false;
 
+    floating(): Array<FloatRule> {
+        return DEFAULT_RULES.concat(this.float)
+    }
+
     window_shall_float(wclass: string, title: string): boolean {
-        return this.float.find((rule) => {
+        return this.floating().find((rule) => {
             if (rule.class) {
                 if (!new RegExp(rule.class).test(wclass)) {
                     return false;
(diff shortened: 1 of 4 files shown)


WatchMkr:
@mmstick is it possible to add a float application window option to g-c-c > Applications instead? I'm leery of introducing a pattern of launching configuration applications from the panel. It gives the feeling of an extension to the operating system rather than an integrated component.

mmstick:
@WatchMkr It's not because there's no association between applications and their windows. `Meta.Window` doesn't have any information about what `.desktop` file opened it.

jacobgkau (changes requested):
The "title" and "class" boxes (or headers) seem to be reversed from the entries in the config file:

shell-pr580-screenshot1

That's what I'm requesting changes for.

Also, do we need any UX input for the dialog? It seems like there's a lot of spacing between the headers at the top, the + sign doesn't have any sort of hover effect like a normal button would (neither do the trash can icons), and the + sign seems awfully close to the text boxes above it.

Do we want to have a checkbox (or on/off switch) to enable/disable log-on-focus? Right now, I still had to manually edit the config file to toggle log-on-focus to see what class I needed to use. Are we expecting people to generally have that information without needing to use log-on-focus? (It would be even better if we could click a window to populate its class and/or title into the dialog, eliminating the terminal from this feature's flow altogether, but not sure what's possible.)

mmstick:
There's also the matter if we could tell what .desktop a window is associated with, we probably don't want every window from that application to float.

mmstick:
Yes, UX input from @maria-komarova  would be good. Although I think our shell theme may be choosing that spacing.

WatchMkr:
@mmstick how about an Always Floating option in the context menu when right-clicking a window?

mmstick:
Might be possible. Looked into it before but I couldn't figure out how. Anyone know of any gnome shell extensions that add context menu options?

WatchMkr:
> There's also the matter if we could tell what .desktop a window is associated with, we probably don't want every window from that application to float.

Do we have an override list that I can review?

mmstick:
Defaults are here: https://github.com/pop-os/shell/blob/master_focal/src/config.ts#L18

PR #589, opened 2020-10-02
improv(rebuild.sh): offer to install TypeScript with npm if needed

toreanderson:
This should make it easier for users of distributions with too old TypeScript versions (e.g., Fedora 32) to try Pop Shell.

This is a second attempt at this change. The first one (#569) was reverted because it did not work with the minimal `sh` found on Debian systems. In this updated version, the `<<<` redirection operator has been replaced with a standard `echo … | …` pipeline. This has been tested to work correctly in a minimal Debian container, using both `sh` and `bash`.

### DIFF
--- a/rebuild.sh
+++ b/rebuild.sh
@@ -13,6 +13,20 @@ else
     echo "Shortcut change already confirmed"
 fi
 
+# Ensure TypeScript is new enough, offer to install with npm if not
+export PATH="./node_modules/.bin:$PATH"
+REQ_TSC_VER="3.8"
+CUR_TSC_VER=$(tsc --version 2>/dev/null || echo 0)
+CUR_TSC_VER=${CUR_TSC_VER##* }
+if [ "$(/bin/echo -e "${REQ_TSC_VER}\n${CUR_TSC_VER}" | sort -V | head -n1)" != "$REQ_TSC_VER" ]; then
+    read -p "TypeScript is not installed or is too old. Install locally using npm? (y/n) " CONT
+    if [ "$CONT" = "y" ]; then
+        npm install typescript
+    else
+        echo "OK. Attempting to continue, but except that the build will fail."
+    fi
+fi
+
 set -xe
 
 # Build and install extension
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ debian/*
 !debian/rules
 .confirm_shortcut_change
 .vscode
+node_modules
+package-lock.json
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ There are additionally those who do want the traditional stacking window managem
 To install this GNOME Shell extension, you MUST have the following:
 
 - GNOME Shell 3.36
-- TypeScript 3.8
+- TypeScript 3.8 (or `npm`, which `rebuild.sh` will use to install TypeScript if necessary)
 - GNU Make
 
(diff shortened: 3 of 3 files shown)


rebuild.sh
```diff
@@ -13,6 +13,20 @@ else
     echo "Shortcut change already confirmed"
 fi
 
+# Ensure TypeScript is new enough, offer to install with npm if not
+export PATH="./node_modules/.bin:$PATH"
```
mmstick, line 17: Do we want to install this in the project folder? Shouldn't it be installed user-wide?

rebuild.sh
```diff
@@ -13,6 +13,20 @@ else
...
+REQ_TSC_VER="3.8"
+CUR_TSC_VER=$(tsc --version 2>/dev/null || echo 0)
+CUR_TSC_VER=${CUR_TSC_VER##* }
+if [ "$(/bin/echo -e "${REQ_TSC_VER}\n${CUR_TSC_VER}" | sort -V | head -n1)" != "$REQ_TSC_VER" ]; then
```
jmmaranan, line 21: I think this will still always ask to install if the version is greater than `tsc 3.8`

rebuild.sh
jmmaranan: I think it depends on the `prefix` (.e.g /usr/bin/node) for `npm install -g`. `npm install` will guarantee that the user will not need sudo.

rebuild.sh
toreanderson: It is supposed to do the opposite, and that's how it works when I've tested it.

Could you run the script under `bash -x` and paste the output up until the point where it asks, please? Also give output from `tsc --version`.

rebuild.sh
toreanderson: @mmstick I personally think that it is the cleanest to install it in the project where it can be trivially cleaned up with `git clean` once Pop Shell has been successfully installed. This is also the default behaviour of `npm install`, so it is the expected behaviour.

If we instead install it somewhere else within the user's `$HOME` means we're polluting it in perpetuity. A global `npm -g`install would, as @jmmaranan points out, require superuser privileges so I consider that approach out of the question.

rebuild.sh
jmmaranan: My bad, I just tested your branch and it seems to work fine using `@latest, @next, @rc, @beta` , etc.

jmmaranan: It looks like the oldest `3.8` is `3.8.3` - that seems to work fine too. Might want to update README.md#developers to have nodejs as a dependency for npm.

toreanderson: The minimum version requirement of `3.8` was blindly copied from [README.md](https://github.com/pop-os/shell/blame/master_focal/README.md#L69) - I haven't verified that this information is correct. I could certainly update this to `3.8.3` if that is the *actual* minimum version requirement of course, but in that case, `README.md` should be updated as well. I feel that is outside the scope of this PR, though.

toreanderson: I've now added a comment to `README.md` stating that `npm` will be used to satisfy the TypeScript dependency if necessary.

jmmaranan:
Hi @toreanderson - https://github.com/pop-os/shell/pull/595 - was added, you might want to move the npm install logic on the folder `scripts`. 
The new dev and local install workflows are `make local-install` and `make debug` respectively.

PR #590, opened 2020-10-02
Integrate Pop-Shell Launcher with external scripts and programs.

jenschurchill:
I really wanted to be able to use shellscript to integrate into the launcher, so I took a crack at creating an external launcher extension.

It is pretty well described in the EXTERNALLAUNCHER.md, but in short, I have added a . (dot) prefix, that examines the following word for a matching script or program in $HOME/.pop-shell.launcher/*

If found, it sends list and apply signals to the script, and presents the result.

This is my first try at GJS (and TypeScript), so I am of course fully open to suggestions.
I would really like to be able to integrate with your preview function, and also present proper icons.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ debian/*
 !debian/rules
 .confirm_shortcut_change
 .vscode
+node_modules
--- a/EXTERNALLAUNCHER.md
+++ b/EXTERNALLAUNCHER.md
@@ -22,14 +22,17 @@ action="${2^^}"
 item_value="${3}"
 
 if [ ! -f "/tmp/bt.idx" ] || [ "$tag" != "$(cat /tmp/bt.idx)" ]; then
+    # New launcher opened, rebuild search index...
     echo "$tag">"/tmp/bt.idx"
     brotab index
 fi
 
 if [ "${action}" == "SEARCH" ]; then
     if [ -n "${item_value}" ]; then
-        brotab search "${item_value}*"
+        # Search tabs in browser, using query...
+        brotab search "${item_value}"
     else
+        # No search query yet provided, list current tabs in browser...
         active_ids="$(brotab active | cut -f1 | tr '\n' '|' | sed 's/\./\\./g')"
         regex="^(${active_ids::-1})\t"
         brotab list | grep -P "$regex"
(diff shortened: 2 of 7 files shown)


mmstick:
My only concern is that if we commit to this particular format, we have to commit to it henceforth forever more. I think this will need some discussion to come to a consensus of what kind of extension API we want to have, and if there are runner APIs that we can adopt to tap into existing runner communities.

mmstick:
I think it'd be good if we could get support for the extension API this uses: https://ulauncher.io/

mmstick:
I guess the main problem with Ulauncher is that's only available as a Python API. I'd rather see some sort of language-agnostic IPC

jenschurchill:
I’ll have a look at ulaunchers API.
I used ulauncher once, but didn’t like it, but that is not to say that the api isn’t alright.

I meant to write earlier, that I am already rearchitecting the code anyway.

I really like the flexibility this affords me, but have come to realise that having the script return the whole dataset once, and caching/filtering it in the extension code isn’t the way forward.

I really want to use my regular toolset, such as fzf, awk, ripgrep, sql etc. So I am changing code to have search itself moved out to the script, instead of the more constrained ‘list’ keyword. And I would prefer a platform agnostic exchange format, like json or the like.

I already had that in mind to begin with, which is why I exposed the session Id, so the script can decide if it wants the cache, index, etc. But I got impatient, and wanted something quick to get the feeling of how my workflow would improve, and I must say, it makes a great difference.

I’ve made some things for rofi as a longtime i3 user, so that might also be something to look at, but for now I’ll change this to a more generic use case, and start researching how to best support a broad integration surface.

Your input is greatly appreciated, thank you.

mmstick:
All we really need from an extension is the icon, name, description, and an ID for that particular choice. Then all the extension needs is to handle lookups, track registered options, and apply options when given by ID.

jenschurchill:
I've found a little time to rework the code, to defer searching to the external script.
Next I'll change to exchange format from tab delimited to JSON.

jenschurchill:
@mmstick : Could you comment on a way, to call the preview/focus code, from inside an extension?

mmstick:
We could have an event in our protocol for triggering this

PR #596, opened 2020-10-04
feat: apply hint color on tabs

jmmaranan:
Use the current active hint logic to apply to the stacking button tab

### DIFF
--- a/src/mod.d.ts
+++ b/src/mod.d.ts
@@ -263,6 +263,7 @@ declare namespace St {
         hide(): void;
         remove_style_class_name(name: string): void;
         remove_style_pseudo_class(name: string): void
+        set_style(inlinecss: string): boolean;
         set_style_class_name(name: string): void;
         set_style_pseudo_class(name: string): void;
         show_all(): void;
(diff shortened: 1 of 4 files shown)


jacobgkau:
This seems to be working fairly well. One potential concern is the text in the tabs being unreadable with dark colors. As an extreme example, take a solid black active hint:

shell-pr596-screenshot1

The active hint is subtle but still visible (some users might prefer this to a bright, highlighted color), and the tab bar is obviously unreadable. Would it be possible to switch the text color in the tabs from black to white if the lightness/luminance of the hint color is below a certain threshold? Something like what's discussed here: https://stackoverflow.com/questions/3942878/how-to-decide-font-color-in-white-or-black-depending-on-background-color

jmmaranan:
Sure, I think that's doable. Thanks for the formula link. I'll check it out

jmmaranan:
@jacobgkau - this is ready for test - the link to the advanced computation of contrast works well

mmstick (approved):

jacobgkau (approved):
Looks great!

PR #601, opened 2020-10-05
fix: Support static workspaces with workspace management in Pop Shell

### LINKED ISSUE #599: Gnome tweaks workspace setting
In gnome tweaks you can chose to have workspaces dynamically created as needed, or to have a fixed number of workspaces. Pop shell's behaviour does not respect this setting and creates a weird mix between the two when you tell tweaks to use static workspaces. 

Let's say you have four fixed workspaces. Whenever a window gets moved to another workspace all is well, unless the window is on the very first or the very last workspace. If it is on the first workspace and you try to move it 'up', so to a non existing zeroth workspace, a zeroth workspace will be created on the fly and your window will get moved to this workspace. It's the same when the window is on the last workspace and you try to move it to the workspace below. 

This behaviour would be sort of welcome in case you want some more space without having to go into the settings, but the drawback is that if you do this by accident you can easily end up with more than 30 workspaces and even freezing a modern laptop. 

I'd like to see any 'dynamically' created workspaces be removed once they're no longer in use, but I'd be more than fine with the behaviour simply reflecting your settings, so no dynamic workspace creation at all. That would make the most sense as well.

mmstick:
Closes #599

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -956,23 +956,31 @@ export class Ext extends Ecs.System<ExtEvent> {
             if (neighbor && neighbor.index() !== ws.index()) {
                 move_to_neighbor(neighbor);
             } else if (direction === Meta.MotionDirection.DOWN && !last_window()) {
-                neighbor = wom.append_new_workspace(false, global.get_current_time());
+                if (this.settings.dynamic_workspaces()) {
+                    neighbor = wom.append_new_workspace(false, global.get_current_time());
+                } else {
+                    return;
+                }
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jacobgkau (approved):
Confirmed this fixes being able to create new workspaces while the setting is set to static. No regressions. [shell-pr601.txt](https://github.com/pop-os/shell/files/5330330/shell-pr601.txt)

Discovered https://github.com/pop-os/shell/issues/602 while testing this PR; it's not a regression in this branch.

PR #607, opened 2020-10-07
feat: make the hint appear much smoother

jmmaranan:
Since the shell is now putting the windows back into their positions pre-tiling, the hint seems to look like being dragged. Need to hide the hint when it is being toggled, or when the tweener is being activated - so the experience is more snappy and smoother.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1612,6 +1612,7 @@ export class Ext extends Ecs.System<ExtEvent> {
         }
 
         const original = this.active_workspace();
+        this.hide_all_borders();
 
         let tiler = new auto_tiler.AutoTiler(
             new Forest.Forest()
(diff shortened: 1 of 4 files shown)


jmmaranan:
When the window came from another monitor, the hint dragging is still noticeable. Finding a fix for that.

jmmaranan:
Fixed the delay caused by the hide all borders loop.

mmstick (approved):

leviport (approved):

mmstick:
@jmmaranan This needs a tiny rebase of one conflict, and it's ready to merge to dev

jmmaranan:
Working on it

jmmaranan:
@mmstick - done rebase

PR #611, opened 2020-10-07
use the hint color for overlay

jmmaranan:
I thought it would be nice to have consistency on the accent. Including the overlay.
Would probably be overkill but here goes:

light colored borders, will use .3 opacity with the same color else
for dark colored borders, default to very light blue hint

image

image


leviport (approved):
This is awesome, thanks!

PR #618, opened 2020-10-08
fix flicker tween observer as near realtime

### LINKED ISSUE #547: Double update / delay

jmmaranan:
Closes #547 

This should fix the flicker due to the higher delay of the tween observer under `on_window_tweened`.
Disabling the whole thing will be worked on a future PR.

### DIFF
--- a/src/tweener.ts
+++ b/src/tweener.ts
@@ -15,6 +15,7 @@ export function add(win: ShellWindow, p: TweenParams) {
     let a = win.meta.get_compositor_private();
     if (!p.mode) p.mode = Clutter.AnimationMode.LINEAR;
     if (a) {
+        remove(a);
         win.hide_border();
         win.update_border_layout();
         a.ease(p);
@@ -35,10 +36,21 @@ export function is_tweening(a: Clutter.Actor) {
 export function on_window_tweened(win: ShellWindow, callback: () => void): SignalID {
     win.update_border_layout();
     win.hide_border();
-    return GLib.timeout_add(GLib.PRIORITY_DEFAULT, 100, () => {
+
+    // the amount of delay for recursively checking
+    // if the animation is done.
+    const tween_timeout = 20;
+    
+    return GLib.timeout_add(GLib.PRIORITY_DEFAULT, tween_timeout, () => {
         const actor = win.meta.get_compositor_private();
-        if (actor && is_tweening(actor)) return true;
-        callback();
+        if (actor) {
+            if (is_tweening(actor)) {
+                return true;
+            } else {
+                remove(actor);
+                callback();
+            }
+        }
         return false;
     });
 }


jacobgkau:
This does visibly improve the opening of new windows. It also significantly slows down the movement of windows in tiling mode, to the point where it's borderline sluggish (although I think we could get away with it.) It looks like we're increasing the animation time from 49ms to 200ms, about a 4x decrease in speed.

This line was just added earlier this week: https://github.com/pop-os/shell/commit/948655b0724a1587b3acfa5f9412ad51d037b0ed 

`const duration = ext.tiler.moving ? 49 : 149;`

Before that line was added, it looks like the value was always set to 100:

`Tweener.add(actor, { x, y, duration: 100, mode: null });`

Is this as simple as longer animation -> less flickering, or would it be possible to bring the animation speed back down to the 100 it was at before while retaining the flicker improvements? Testing locally, it looks like the animation speed shouldn't matter-- I'm able to decrease from 200 to the old 100, or even the current 49, and I still see an improvement over master. I'm still seeing occasional flickers, but I also see them occasionally when testing this branch unmodified.

jmmaranan:
@jacobgkau - good catch on the 200ms - that was not intentional. 49ms should still be the value.

I am curious on the flicker that you are seeing too. I do not see the windows going back to where it started, after it calls `onComplete()`. Are you referring to the first time a window paints? And then after shell grabs it for animation?

jmmaranan:
I have reverted 200ms to 49ms.

jacobgkau (approved):
> I am curious on the flicker that you are seeing too. I do not see the windows going back to where it started, after it calls `onComplete()`. Are you referring to the first time a window paints? And then after shell grabs it for animation?

After the shell has started moving the window toward its new tiled position, it flickers back to its original position. This was happening basically every time before this PR, and it's much more occasional after this PR.

I managed to catch it on a screen recording. This was recorded on the serw12 (with a 144Hz panel): https://www.dropbox.com/s/m53betzf1197trr/shell-pr618-screencast1.mkv?dl=0 

In the recording, it happens the second time I open gnome-terminal, and also the last time. Interestingly, it didn't seem to happen on the serw12 while I was recording with GNOME's built-in screen recorder, which is why I resorted to SimpleScreenRecorder. I did also see it happen on the galp4 (Intel graphics & 60Hz panel) with the GNOME screen recorder: https://www.dropbox.com/s/0edh70utlzhazhf/shell-pr618-screencast2.webm?dl=0 

Since this is an improvement and we've addressed the animation time, I'm approving to merge to dev. I can re-review if there are any more improvements to be made here, otherwise they can go in future PRs.

PR #628, opened 2020-10-13
Floating Window Exceptions GUI Support

### LINKED ISSUE #513: GUI for adding windows to the floating list

### LINKED ISSUE #622: Mockups for the floating window exceptions feature
This is a feature to help improve auto-tiling experience by having a list of exceptions to always open in floating mode. 

Because we can only rely on application class and title, and because users generally feel the need to add an application or a window to the exceptions when they encounter it, it seems the best way would be to enable people adding an exception by selecting an open window on the current workspace. 
Selecting "Floating Window Exceptions" in the tiling menu would open a window that lists both personal user exceptions and system defaults that are updated as we get more user reports and validate them. 
extension-menu-1
floating-exceptions-1
floating exceptions-2

Selection mode is similar to the window preview in GNOME Shell Overview: 
floating-selection-mode

System dialog that shows after a user clicks on the window offers an option to add all windows of an application or only the current one to the exceptions. 
system-dialog-add-page

The basic prototype shows the flow for adding Slack to the exception: https://www.figma.com/proto/uI6TbZgpvBwKDN6DO9VxX5/Pop-_OS-various?node-id=78%3A324&scaling=scale-down

mmstick:
Closes #513 
Closes #622


jacobgkau:
The button wording seems a little off (I see that the wording was in the mockups):

shell-pr628-screenshot1

Instead of "This App Windows", should it be "This App's Windows" or "All App Windows"?

Also, there's currently no check for duplicates, so I can add the same exception to the list over and over again:

shell-pr628-screenshot2

That could either be a bug or not something we're concerned about. However, if I delete one of those duplicate entries, all of them are deleted in the config file, but the rest of them don't disappear from the GUI until I close the list and re-open it. That is an issue.

WatchMkr:
Let's go with "This App's Windows".

mmstick:
I tried to use a `Set<T>` to filter duplicates, but apparently JavaScript's sets aren't that reliable without some additional machinery.

src/floating_exceptions/src/main.ts
```diff
@@ -0,0 +1,313 @@
...
+    layout.attach(icon, 1, 0, 1, 2)
+
+    let button = Gtk.Button.new()
+    button.relief = Gtk.ReliefStyle.NONE;
+    button.add(layout)
+
+    return button
+}
+
+class MainView implements View {
```
jmmaranan, line 81: @mmstick on typescript 4.x, if not prepending `export` the tsc will write `export {}` at bottom and gjs will complain it needs to be at the top. Float will not open

mmstick:
So there's nothing I can do about the close buttons in the overview, since that's part of GNOME Shell's internal behaviors.

I've no idea where I would submit user configs, so I've not added that section. You can't create issues

The other things I can fix

maria-komarova:
So there is no way to do it so that users won't be able to close windows while selecting them for exceptions?

mmstick:
Yeah, it's much easier to add features than it is to remove them.

WatchMkr:
@mmstick we could host something like this collect issue reports: https://github.com/schneidmaster/gitreports.com

We wouldn't want a high frequency. Maybe submit floating lists as issues once a week. And notifications muted. The QA team could then test the issue and submit a PR if warranted.

What do you think?

mmstick:
Yeah that would work

jmbuhr:
Is there some middle ground between "This app's windows" and "Current window only"? I assume the first is based on `WM_CLASS(STRING)`, while the second ist very specific. I often encounter Apps whose supposed popout windows have the same class as the app, so the first method doesn't work because it would float everything, but have a different `WM_NAME(STRING)`. Maybe this could be used?

jacobgkau:
This is looking good. I noticed there's no icon and the name of the Floating Window Exceptions app shows as "Gjs" in the top panel and launcher while it's running. Do we want to set a name and icon for this app? (shell-shortcuts, for example, has an icon and name, as do other utilities such as Repoman and Transition.)

shell-pr628-screenshot3

mmstick:
I set `application-default` as the icon, `pop-shell-exceptions` as the prgname, and `Pop Shell Floating Window Exceptions` as the application name (seen in the overview).

jmbuhr:
For me, when setting a specific window to floating, closing the window and creating a new one of the same type, it originally starts out tiled and only becomes floating after toggling tiling mode off and on with a double tap of Super+Y. But I only tested it with plot windows in RStudio so far.

mmstick:
It sounds like the window is dynamically changing its title after it opens. The decision to tile is decided when the window is opened

jacobgkau:
> I set `application-default` as the icon, `pop-shell-exceptions` as the prgname, and `Pop Shell Floating Window Exceptions` as the application name (seen in the overview).

Very minor, but it looks like the prgname is showing up in the overview and not the application name:

shell-pr628-screenshot6

shell-pr628-screenshot5

shell-pr628-screenshot7

Text wrapping for long window titles is working well, but it does cause a scrollbar to appear even though there's still more space for the list to vertically expand. (If I only add exceptions with short titles/classes, the scrollbar doesn't appear until the list has reached the bottom of the window.)

shell-pr628-screenshot4

I've also noticed that after restarting GNOME Shell with Alt-F2+r+Enter while the Floating Window Exceptions window is open, the "Select" button doesn't work anymore until I close and re-open Floating Window Exceptions. That doesn't seem like too big of a problem, I wouldn't expect people to have to restart the shell while this window's open.

mmstick:
Yeah it won't work anymore because GNOME Shell's not listening to process if you restart it.

mmstick:
The ScrolledWindow becoming scrollable when the ListBox is resized by the labels increasing in size seems to be a limitation of GTK.

WatchMkr:
@maria-komarova @mmstick should we ellipsis instead? I don't think the detailed info, especially if it gets that long, is useful to users. That also avoid different size list boxes that can look bad.

maria-komarova:
> @maria-komarova @mmstick should we ellipsis instead? I don't think the detailed info, especially if it gets that long, is useful to users. That also avoid different size list boxes that can look bad.

Yeah, ellipsis would do fine. It does wrap quite a lot.

maria-komarova:
> It sounds like the window is dynamically changing its title after it opens. The decision to tile is decided when the window is opened

What happens when we toggle window to floating? Could we potentially apply the same logic here when someone adds a window to the exceptions?

mmstick:
Toggling via shortcut sets a floating tag without making any changes to the config, so it's temporary.

jacobgkau:
The list/scrolling behavior looks fine after the change to use ellipses. Adding and removing exceptions seems to be working as intended, and I am also able to disable the built-in exceptions using the on/off switches.

Regression testing passed: [shell-pr628.txt](https://github.com/pop-os/shell/files/5394146/shell-pr628.txt)

One last point on the copy, it seems inconsistent that there's no period at the end of the sentence at the top of the window when there is a period at the end of the sentence a bit lower down:

shell-pr628-screenshot9

I previously thought it was just me being overly picky, but it actually had a period in the mockup: https://github.com/pop-os/shell/issues/622 

After that, this looks good to approve from QA's side.

mmstick:
I'll add that period and merge

jacobgkau (approved):

PR #646, opened 2020-10-20
Add default pointer position config when you switch in another window

bnjjj:
Hello, as a new pop!_os user I tried to configure the default position for my pointer on windows when switching with keyboard shortcuts. As it's my first pull-request for this project let me know if I miss something. Oh and can you tell me how can I test this code without breaking all my customs configuration on my own pop!_os environment ? (without breaking my custom keyboard shortcuts please)

### DIFF
--- a/src/window.ts
+++ b/src/window.ts
@@ -1,19 +1,21 @@
 // @ts-ignore
 const Me = imports.misc.extensionUtils.getCurrentExtension();
 
+import * as Config from 'config';
 import * as lib from 'lib';
 import * as log from 'log';
 import * as once_cell from 'once_cell';
 import * as Rect from 'rectangle';
 import * as Tags from 'tags';
+import * as Tweener from 'tweener';
 import * as utils from 'utils';
 import * as xprop from 'xprop';
-import * as Tweener from 'tweener';
-
 import type { Entity } from './ecs';
 import type { Ext } from './extension';
 import type { Rectangle } from './rectangle';
 
+
+const { DefaultPointerPosition } = Config;
 const { Gdk, Meta, Shell, St, GLib } = imports.gi;
 
 const { OnceCell } = once_cell;
@@ -108,7 +110,7 @@ export class ShellWindow {
     }
 
     activate(): void {
-        activate(this.meta);
+        activate(this.ext.conf.default_pointer_position, this.meta);
     }
 
     actor_exists(): boolean {
@@ -351,7 +353,7 @@ export class ShellWindow {
         let br = other.rect().clone();
 
         other.move(ext, ar);
-        this.move(ext, br, () => place_pointer_on(this.meta));
+        this.move(ext, br, () => place_pointer_on(this.ext.conf.default_pointer_position, this.meta));
     }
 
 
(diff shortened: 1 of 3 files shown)


mmstick:
You can

```
make && make install && make restart-shell
```

If you want to compile, install, and restart the shell without doing any configuration changes.

bnjjj:
Alright, it completely fucked up my pop os :p I don't have pop shell enabled anymore and can't re-enable it :(

bnjjj:
Ok I made a few changes. Now it works like a charm, if you need a video to see the behavior let me know. In fact I also added a configuration named `move_pointer_on_switch` which enable the pointer move also when you switch applications with `meta+tab`, by default it's set to false to keep previous behavior.

mmstick:
@bnjjj There are logging levels you can toggle in extension settings. The default level is `OFF`. The higher the level, the more logs you see.

bnjjj:
Ok got it @mmstick thanks ! Could you review this PR please ? I'm open to any suggestions :) BTW here is a demo to discover how it works https://twitter.com/BnJ25/status/1319977536894521345?s=20

mmstick (approved):

src/window.ts
```diff
@@ -522,25 +524,61 @@ export class ShellWindow {
     private window_raised() {
         this.restack(RESTACK_STATE.RAISED);
         this.show_border();
+        if (this.ext.conf.move_pointer_on_switch && !this.pointer_already_on_window()) {
```
jacobgkau, line 527: Is not moving the pointer if the pointer's already somewhere on the activated window what we really want? On one hand, it prevents unnecessary cursor movement if the user just wants the cursor near the window, but on the other hand, it makes the behavior a little less consistent/predictable.

Video to show what I mean: https://www.dropbox.com/s/rzoeicovtm33ps0/shell-pr646-screencast1.mp4?dl=0 Notice how the cursor is moved to the bottom right of the Files app the first time I switch to it, but not the second time, because the cursor happened to already be on top of the Files app the second time.

Another example of the same inconsistency when switching workspaces (without using Alt-Tab): https://www.dropbox.com/s/eyb5o5zmx9en0rn/shell-pr646-screencast2.mp4?dl=0 

If we're good with this as intended behavior (which is disabled by default anyway), and if "switch" is the correct terminology to use for this option (as opposed to "activation" or "focus"), then this is fine to merge to dev.

jacobgkau (approved):

src/window.ts
mmstick: I'm open to discussions on the behavior. The current behavior always places the pointer on the top left to aid mouse movements on ultrawide displays. The top left corner of an application usually has the most clickable controls, and is generally out of the way

bnjjj: @jacobgkau Yes in fact it's the intended behavior. The reason why I enabled this is because if you switch application just with your mouse by cliking on the window, if I keep behavior to move cursor it becomes very weird because your pointer suddenly moves to another position, which is not expected. As a user of this feature for a week now I have to say I'm pretty satisfy with this behavior and don't feel inconsistency but yes here is my humble opinion. If you have any other suggestions I'm also open to discussion :)

jacobgkau: > if you switch application just with your mouse by cliking on the window, if I keep behavior to move cursor it becomes very weird because your pointer suddenly moves to another position

I agree, we wouldn't want that, so the current behavior's fine. Thanks for the explanation.

PR #649, opened 2020-10-21
Display shortcuts website if pop-shell-shortcuts is not installed

carlwgeorge:


mmstick (approved):

jacobgkau:
This is a great idea! It's working fine for the "View All" menu item. Do we want to do the same thing on [line 144](https://github.com/pop-os/shell/blob/25f353219dba01a82aa098f7f8b8726306548cf0/src/panel_settings.ts#L144) so clicking on the three shortcuts shown in the menu also works? That would be consistent with the behavior when shell-shortcuts is installed.

carlwgeorge:
@jacobgkau Ah, I didn't realize clicking there worked too.  I just updated this to modify that line too.

jacobgkau:
It's failing to build now:

```
system76@pop-os:~/shell$ make
UUID is "pop-shell@system76.com"
depcheck
rm -rf _build schemas/gschemas.compiled target
tsc
src/panel_settings.ts:4:1 - error TS6133: 'log' is declared but its value is never read.
4 import * as log from 'log';
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error.
make: *** [Makefile:25: transpile] Error 2
```

Looks like [this line](https://github.com/pop-os/shell/blob/3f611bce0d2b2ac114ec6bc035912d9de8daec5e/src/panel_settings.ts#L4) needs to be removed or commented out since there are no longer any "log" lines in this file, should be fine after that.

carlwgeorge:
Fixed, thanks for catching that.

jacobgkau (approved):
Thank you!

mmstick (approved):

PR #654, opened 2020-10-23
Don't move always on all workspaces windows

cansozbir:
While creating an upper workspace, the extension adds a new workspace below and push all the non-selected windows down, and in GNOME, moving a window selected as "Always on Visible Workspace" to another workspace disables that option from the window. This pull request aims to solve that issue.

Steps to reproduce the issue:

1.Open window A and window B on the first workspace you have.
2.Right-click on the window title of the window B and select "Always on Visible Workspace"
3.Select window A, and press Super+Shift+Up to create an upper workspace.
4.You will notice that window A is moved to a new workspace, but window B is not "Always on Visible Workspace" anymore.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1294,7 +1294,7 @@ export class Ext extends Ecs.System<ExtEvent> {
         function window_move(ext: Ext, entity: Entity, ws: WorkspaceID) {
             if (change_workspace) {
                 const window = ext.windows.get(entity);
-                if (!window || !window.actor_exists()) return;
+                if (!window || !window.actor_exists() || window.meta.is_on_all_workspaces()) return;
 
                 ext.size_signals_block(window);
                 window.meta.change_workspace_by_index(ws, false);
--- a/src/mod.d.ts
+++ b/src/mod.d.ts
@@ -201,6 +201,7 @@ declare namespace Meta {
         is_above(): boolean;
         is_client_decorated(): boolean;
         is_fullscreen(): boolean;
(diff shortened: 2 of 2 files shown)


mmstick:
Does this work even if you move a window around the tree before moving workspaces?

cansozbir:
> Does this work even if you move a window around the tree before moving workspaces?

I'm not sure about did I understand you very well but.
I tried to move the window on the screen (not between workspaces)  while it is selected as "Always on Visible Workspace". There was no issue for me on both, tiling disabled and enabled. 
Sorry if I got you wrong.

mmstick (approved):

jacobgkau (changes requested):
Failed to build on both Focal and Groovy. Getting this when trying to build it locally on Groovy:

```
system76@pop-os:~/shell$ make
UUID is "pop-shell@system76.com"
depcheck
rm -rf _build schemas/gschemas.compiled target
tsc
src/extension.ts:1297:70 - error TS2339: Property 'is_on_all_workspaces' does not exist on type 'Window'.

1297                 if (!window || !window.actor_exists() || window.meta.is_on_all_workspaces()) return;
                                                                          ~~~~~~~~~~~~~~~~~~~~


Found 1 error.

make: *** [Makefile:25: transpile] Error 2
```

cansozbir:
Sorry, I fixed it. And now you should be able to build it without any problem.

jacobgkau (approved):
This is working now, fine to merge into dev.

PR #658, opened 2020-10-25
Use XDG paths where possible

maxeaubrey:
Minor change to use XDG paths for the configuration file and installing for local development.

Apologies if the commit messages aren't inline with what you use, tried to base them on others, but I wasn't sure how to categorise this change.


mmstick:
Thanks for the contribution!

In case you're wondering where the commit style comes from, we adhere to [Conventional Commits guidelines](https://www.conventionalcommits.org/en/v1.0.0-beta.2/).

PR #668, opened 2020-10-28
feat: Support toggling of tiling for floating exceptions

### LINKED ISSUE #667: Allow tiling a floating-exception window with Super+G
Example use case: Zoom
Zoom has 20 different windows. Identifying and managing windows that should float and tile isn't feasible so we set the entire app to float. However, a user may want to tile the video window while in a conference. Allowing Super+G to override would accomplish this and provide consistent shortcut behavior.

mmstick:
Closes #667


jacobgkau (approved):
Working as expected (able to press Super-G to tile a window that was floating by default due to being configured as an exception; new windows for that app still float by default, and manually-tiled windows still float by default after tiling is disabled/re-enabled.)

Good to merge to dev for further regression testing.

PR #679, opened 2020-11-02
fix: Quitting fullscreen apps will now recover tab visibility

mmstick:


jackpot51 (approved):

jacobgkau (approved):
Confirmed this fixes the described issue (stacking, full-screening, and quitting an app no longer hides the remaining tabs in the stack.) Regression testing passed: [shell-pr679.txt](https://github.com/pop-os/shell/files/5489070/shell-pr679.txt)

PR #683, opened 2020-11-04
fix: Add floating exception for system76-driver

ids1024:
`system76-driver` doesn't support rescaling. I looked at trying to make it resize, but doing it with a layout that doesn't look funny seems difficult. It makes sense for that kind of program to be floating, probably.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -34,6 +34,7 @@ export const DEFAULT_RULES: Array<FloatRule> = [
     { class: "TelegramDesktop", title: "Media viewer" },
     { class: "Slack", title: "Slack | mini panel" },
     { class: "Solaar", },
+    { class: "system76-driver", },
     { class: "zoom", },
 ];
 


jacobgkau (approved):
If we're not going to make it resizable, it makes sense for it to float. (Tested & the class preset works.)

> I looked at trying to make it resize, but doing it with a layout that doesn't look funny seems difficult.

For what it's worth, I'm fairly certain the "Create Log Files" button is what gets the majority of the usage in this app, so we may not be tied to the current layout. I can see how a single-button app scaling up might look funny in any case, though. Might be something @maria-komarova could chime in on.

My only concern with floating would be the driver app not being visible if tiling is enabled and another window keeps grabbing focus or a user's having some sort of input issue that's causing other windows to be selected, but tiling can always be disabled if some weird problem like that requires it.

maria-komarova:
I am fine with adding it to the list of exceptions for now. It is not great if it gets hidden under other windows but there are ways of finding it through the overview etc. 
If I remember correct we were considering doing something about the driver. I am not sure where we are with the plan but I will check with everyone and try to figure out what needs to be done about it in terms of design.

PR #701, opened 2020-11-10
fix: Improve handling of tab-dragged windows from browsers

### LINKED ISSUE #673: chrome tab dragging buggy
chrome tabs rapidly jumping from the starting point of drag to the position of the mouse several times a second

drag an individual chrome tab

Pop!_OS 20.10

**(5) Gnome Shell version:**
3.38.1

```1.0.0~1603994289~20.10~ecc9b38```

pop-os default

1x 1080p

mmstick:
Finally thought of a way to improve this. It isn't perfect, but it's significantly better than the current behavior.

Closes #673

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -144,6 +144,8 @@ export class Ext extends Ecs.System<ExtEvent> {
     /** Record of misc. global objects and their attached signals */
     private signals: Map<GObject.Object, Array<SignalID>> = new Map();
 
+    private size_requests: Map<GObject.Object, SignalID> = new Map();
+
     /** Used to debounce on_focus triggers */
     private focus_trigger: null | SignalID = null;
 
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Definitely not perfect, but an improvement over current behavior for sure. The window still won't move within the tree or between displays when dragged by the tab, but tabs can be dragged between windows without the visual glitch now.

Regression testing passed. I briefly had some trouble resizing stacks with the mouse, but after switching to stable and then back to this branch, I'm not able to replicate that issue anymore. (I am sometimes unable to resize a stacked Chromium window by its top edge while not using system titlebars, but that also happens in the stable version.) [shell-pr701.txt](https://github.com/pop-os/shell/files/5520485/shell-pr701.txt)

PR #704, opened 2020-11-10
Add pop-shell-exceptions to system exceptions

### LINKED ISSUE #690: Adding 'pop-shell-exceptions' to floating window exceptions
This is more of a feature request, than a bug (tried using chat.pop-os.org but got stuck with "No teams are available to join").

My proposition is to add the `pop-shell-exceptions` (the exception chooser) WM_CLASS to floating window exceptions.
I haven't found a way yet to add it by myself since it only shows other windows and doesn't allow to manually specify a WM_CLASS.

erazemk:
This commit adds the "Floating Window Exceptions" window on Wayland to system exceptions.
This is needed because while it is floating by default on Xorg, it is not on Wayland.

Closes #690.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -36,6 +36,7 @@ export const DEFAULT_RULES: Array<FloatRule> = [
     { class: "Solaar", },
     { class: "system76-driver", },
     { class: "zoom", },
+    { class: "pop-shell-exceptions", },
 ];
 
 export interface FloatRule {


jacobgkau (commented):
Confirmed this window doesn't float by default in Wayland before this change, and does float by default after the change.

In Xorg, the window already floated by default. The one problem with adding it to the list like this is that flipping the on/off switch in the exception list to "Off" in Xorg doesn't un-float the window like a user would expect:

shell-pr704-screenshot1

It continues floating even though the exception has been turned off.

@mmstick Since it's supposed to float by default as a dialog window, do you think it would be better to fix the Wayland behavior so dialogs float by default, or is this workaround the best option? If we add this to the exceptions list, should we make it not a dialog window so the on/off switch works properly?

mmstick:
It's better to find out why a GTK Dialig is floating in X11, but not Wayland.

erazemk:
@mmstick Just checked and on Wayland the wmclass is `pop-shell-extensions`, but on xorg it's `Floating Window Exceptions`, so the problem is that it has different wmclasses.

mmstick:
The value of the wmclass shouldn't matter, since only normal windows are allowed to tile. Dialog windows are not.

PR #705, opened 2020-11-10
fix: Improvements to display hotplugging

mmstick:
- Improves how forks are moved when unplugging a display
- Also improves handling of monitors that disappear only briefly (such as toggling power)

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -93,6 +93,9 @@ export class Ext extends Ecs.System<ExtEvent> {
     /** The currently-loaded theme variant */
     current_style: Style = this.settings.is_dark_shell() ? Style.Dark : Style.Light;
 
+    /** Set when the display configuration has been triggered for execution */
+    displays_updating: SignalID | null = null;
+
     /** Row size in snap-to-grid */
     row_size: number = 32;
 
@@ -520,6 +523,25 @@ export class Ext extends Ecs.System<ExtEvent> {
         return [primary, this.displays.get(primary) as Display];
     }
 
+    find_unused_workspace(): [number, any] {
+        if (!this.auto_tiler) return [0, wom.get_workspace_by_index(0)]
+
+        let id = 0
+
+        for (const fork of this.auto_tiler.forest.forks.values()) {
+            if (id < fork.monitor) id = fork.monitor
+        }
+
+        let new_work
+        if (id + 1 === wom.get_n_workspaces()) {
+            new_work = wom.append_new_workspace(true, global.get_current_time())
+        } else {
+            new_work = wom.get_workspace_by_index(id + 1)
+        }
+
+        return [id + 1, new_work];
+    }
+
     focus_window(): Window.ShellWindow | null {
         let focused = this.get_window(display.get_focus_window())
         if (!focused && this.last_focused) {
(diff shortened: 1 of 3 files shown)


jacobgkau:
After hotplugging a display, I'm not able to move windows onto that display with the keyboard or mouse: https://www.dropbox.com/s/fh3wmkhlu4sd8ds/shell-pr705-screencast1.mkv?dl=0 

After restarting GNOME Shell, I'm able to move windows onto the second display. When I unplug the second display, the relocated windows do move to their own workspace now (which is an improvement), but they still appear off of the screen. For example, if I start with this layout (no other windows/workspaces present):

shell-pr705-before

After unplugging the second display, these are the two workspaces I have on the remaining display:

shell-pr705-after1

shell-pr705-after2

Where the second one (which is actually on top, but not activated by default) is not correct considering tiling is still enabled.

mmstick:
What's your display configuration? Haven't been able to replicate these issues on my dual monitor setup.

jacobgkau:
The display configuration is an internal 1080p display (primary) and an external 1440p display (secondary) that I am plugging/unplugging.

I've found another issue. My testing yesterday was with the top edges of the displays lined up (as shown in the screenshot.) If I line up the bottom edges instead, I get windows on the primary display sticking up off of the screen. Before plugging in:

Screenshot from 2020-11-12 09-24-29

After plugging in (anything above the top panel is not physically visible):

Screenshot from 2020-11-12 09-24-50

Interestingly, on the serw12, I am able to move the windows to the second display when plugging in via Mini DisplayPort or HDMI, but not when plugging in via USB-C. No clue why the physical port would make a difference.

Attempting to move a window to a hot-plugged display with USB-C: https://www.dropbox.com/s/a65qu4abuvrkw36/shell-pr705-screencast2.mp4?dl=0 

Attempting to move a window to a hot-plugged display with Mini DisplayPort: https://www.dropbox.com/s/h5v9rzsb4264hsn/shell-pr705-screencast3.mp4?dl=0 

If I close and re-open the windows while the second display is plugged in (which brings them into the normal viewing area), they hang off the bottom of the internal display once the external display is unplugged:

Screenshot from 2020-11-12 09-59-51

On the galp5, I'm seeing the same behavior as yesterday (with HDMI or USB-C, any graphics mode): no hanging windows as long as the displays' top edges are lining up, but can't move them to the second display (this happens with two, three, or four windows, but not with one window.) I get the same hanging windows as soon as I modify the configuration so the top edges are no longer lining up.

> Unplugging the primary left monitor moves everything to the right monitor. Plugging it back in, windows still move to it via keyboard and mouse.

Make sure to restart GNOME Shell after unplugging the second display, before plugging it back in. Right now, I'm going into this as if I booted with the laptop screen only and am plugging into a second display.

jacobgkau:
This is working better now.

- After plugging in a secondary display for the first time, all workspaces that were previously on the single display are moved onto the new display. Is this intentional?
  - `TESTING.md` says "Plug an additional display into a laptop - windows and workspaces don't change." We would need to modify that test if we're expecting all windows to automatically move to a new display when it's plugged in. (I'm not sure this would always make sense, e.g. in situations where a laptop is being connected to a projector for a presentation.)
- While the trees now usually adjust themselves a second or two after the display configuration's been changed, when the top edges of the displays don't line up, I still have trees on the remaining display hanging off the bottom:
shell-pr705-screenshot7
To clarify, trees that were on the external display that I've unplugged appear correctly; trees that were already on the remaining internal display start hanging after I unplug the external display. This seems more likely to happen if I'm utilizing multiple workspaces before unplugging the second display.

- Related to the first point, if I re-plug the external display while I have these trees hanging off the bottom edge, any trees that were hanging are _not_ moved onto the external display.

PR #708, opened 2020-11-11
Override terminal launch shortcut to be in line with others

Xanewok:
Before this change, terminal seemed to be the only shortcut not being overridden (at least on default Ubuntu 20.10) from the Keyboard Shortcuts > Launchers section and used Ctrl+Alt+T instead of Super+T.

This change aims to make things a bit more consistent and hopefully facilitate transition from other desktop environments (like Xfce).

Tested with `make local-install` and seems to be working correctly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ These are key to many of the shortcuts utilized by tiling window managers. This 
 - `Super` + `f`: Files
 - `Super` + `e`: Email
 - `Super` + `b`: Web Browser
+- `Super` + `t`: Terminal
 
 ### Window Management Mode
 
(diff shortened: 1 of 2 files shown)


Xanewok:
Was this PR closed because the target branch `master_groovy` was renamed or is this something you're not considering at all?

Great work by the way, it provides great tiling window experience so far and is something I've been missing in gnome :)

mmstick:
Removal of the branch automatically closed this

Xanewok:
Opened #722 since it seems I can't change the base branch of the PR (maybe you can do that only for repos you have write access despite being the PR author? idk :shrug:), I hope that's okay.

PR #711, opened 2020-11-12
fix: Unminimized windows attaching incorrectly to the tree

### LINKED ISSUE #706: Ghost window after minimizing & un-minimizing window within a stack
**(1) Issue/Bug Description:** After minimizing a stacked window and then restoring that window, a ghost window appears. (This also causes the stack to move together with windows near the stack.)

1. Stack some windows.
2. Minimize one of the windows in the stack.
3. Attempt to move some windows around.

**(3) Expected behavior:** No ghost window.

**(5) Gnome Shell version:** 3.38.1

- `04eaf4f` (current master_groovy)
- Discovered while testing `90a07ab` on primary_groovy branch

**(7) Where was Pop Shell installed from:** The master staging repo.

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):** 1x internal 1080p, 1x external 1440p.

**(9) Other Installed/Enabled Extensions:** Pop!_OS defaults only. (Dash to Panel is installed but not enabled.)

**(10) Other Notes:** Video: https://www.dropbox.com/s/ikhhspajfgnz3tw/simplescreenrecorder-2020-11-10_16.48.44.mp4?dl=0

mmstick:
Closes #706

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -723,7 +723,8 @@ export class Ext extends Ecs.System<ExtEvent> {
                 + `  name: ${win.name(this)},\n`
                 + `  rect: ${win.rect().fmt()},\n`
                 + `  workspace: ${win.workspace_id()},\n`
-                + `  xid: ${win.xid()},\n`;
+                + `  xid: ${win.xid()},\n`
+                + `  stack: ${win.stack},\n`
 
             if (this.auto_tiler) {
                 msg += `  fork: (${this.auto_tiler.attached.get(win.entity)}),\n`;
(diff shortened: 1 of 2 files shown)


jacobgkau:
This does fix #706. If I minimize a window from a stack, un-stack the remaining window(s), then un-minimize the first window, it's not tiled until I move it again: https://www.dropbox.com/s/80lv96ub7rale1c/shell-pr711-screencast1.mkv?dl=0

This doesn't happen on master (when I try to replicate this new issue on master, the un-minimized window is tiled before the ghost window appears.) With this PR, it looks like the window will always try to un-minimize into its stack; if that's the case, then I think it will need to check if its stack still exists somewhere in the process, and un-minimize into the tree normally if the stack doesn't exist.

mmstick:
Fixed

jacobgkau (approved):
Looks good now, issue is fixed & no regressions. [shell-pr711.txt](https://github.com/pop-os/shell/files/5533420/shell-pr711.txt)

I noticed that if I minimize a window, then navigate to the edge of the screen and then try to navigate further (e.g. navigating to the bottom and then pressing `Super+Down` again), that can un-minimize the window. It's not always the same edge of the screen that brings the minimized window back. Is that expected behavior, or should I open an issue to track it? (It's also happening in master.)

mmstick:
Open an issue so we can track it. Sounds strange

PR #717, opened 2020-11-13
chore(exceptions): Add IntelliJ WebStorm exceptions

mmstick:


jacobgkau (approved):
"Welcome to WebStorm" window doesn't grow, and definitely needs to be floated. "License Activation" and "Customize WebStorm" are resizable; "Customize WebStorm" has a minimum size that could be problematic. Not sure "License Activation" needs to be floated, but it's essentially a dialog, so if we want to float it, that's fine.

How the windows looked when tiled:

Screenshot from 2020-11-13 10-02-37

Screenshot from 2020-11-13 09-58-54

Screenshot from 2020-11-13 09-57-31

All of the exceptions work as expected.

PR #719, opened 2020-11-13
Window being selected too soon from exceptions select

mmstick:
Sometimes the exceptions dialog selects a window immediately after the overview opens

### DIFF
--- a/src/dialog_add_exception.ts
+++ b/src/dialog_add_exception.ts
@@ -61,7 +61,7 @@ export class AddExceptionDialog {
     }
 
     show() {
-        this.dialog.show_all();
+        this.dialog.show();
     }
 
     open() {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -46,6 +46,7 @@ const Tags = Me.imports.tags;
 
 const STYLESHEET_PATHS = ['light', 'dark'].map(stylesheet_path);
 const STYLESHEETS = STYLESHEET_PATHS.map((path) => Gio.File.new_for_path(path));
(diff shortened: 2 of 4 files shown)


jacobgkau:
On `763b699`, I'm seeing cases where I don't get the confirmation dialog after clicking on a window; instead, it waits until I change focus to a second window before popping up. https://www.dropbox.com/s/ur638mvaitous76/shell-pr719-screencast1.mkv?dl=0

jacobgkau:
On Groovy, exception selection is working normally again, and regression testing passed. [shell-pr719-groovy.txt](https://github.com/pop-os/shell/files/5632670/shell-pr719-groovy.txt)

On Focal, I've had multiple instances where the confirmation dialog shows up before I've chosen a window:

Screenshot from 2020-12-02 16-43-08

Screenshot from 2020-12-02 16-47-23

It happens fairly reliably if I close everything, restart GNOME Shell, open a couple of windows, and try to select an exception. I went back to Groovy and am still not seeing it happen there.

jacobgkau (approved):
Not seeing any more early-firing of the exceptions dialog. Regression testing passed (I'm seeing some behavior with terminal & Nautilus for "current window only" that I don't remember, but I've confirmed that happens on current stable as well. https://github.com/pop-os/shell/issues/778)

[shell-pr719-focal-run2.txt](https://github.com/pop-os/shell/files/5698906/shell-pr719-focal-run2.txt)
[shell-pr719-groovy-run2.txt](https://github.com/pop-os/shell/files/5698907/shell-pr719-groovy-run2.txt)

PR #710, opened 2020-11-12
fix: Moving tiled windows between monitors in overview

mmstick:
Closes #633

PR #720, opened 2020-11-13
fix: Moving tiled windows between monitors in overview

### LINKED ISSUE #633: In tiling mode, can't move window between monitors from overview mode
* Connect more than one monitor
* Enable "Tile Windows"
* Press `super` to open overview mode
* Try to drag window from one monitor to the other

**(3) Expected behavior:**

The window should be moved the the second display, but Pop!_Shell seems to move it back.

This works when tiling mode is off, or when not in overview mode. It also is able to move a window to another virtual desktop in overview mode without issue.

Pop!_OS 20.04

**(5) Gnome Shell version:**

`3.36.4-1ubuntu1~20.04.2pop1~1597874448~20.04~260a1a7`

`0.1.0~1602687352~20.04~fd2380c`

mmstick:
Closes #633

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -323,11 +323,31 @@ export class AutoTiler {
      * - If the window is dropped onto a window, tile onto it
      * - If no window is present, tile onto the monitor
      */
-    on_drop(ext: Ext, win: ShellWindow) {
-        if (this.dropped_on_sibling(ext, win.entity)) return;
-
+    on_drop(ext: Ext, win: ShellWindow, via_overview: boolean = false) {
         const [cursor, monitor] = ext.cursor_status();
         const workspace = ext.active_workspace();
+
+        const attach_mon = () => {
+            const workspace_id: [number, number] = [monitor, workspace]
+            const toplevel = this.forest.find_toplevel(workspace_id);
+            if (toplevel) {
+                const attach_to = this.forest.largest_window_on(ext, toplevel);
+                if (attach_to) {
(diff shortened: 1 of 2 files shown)


mmstick:
Found the cause. Revised fix in a minute

jacobgkau (changes requested):
The ghost window issue described in [this comment](https://github.com/pop-os/shell/pull/710#pullrequestreview-529391221) is back in `0a7d4db`.

mmstick:
I see. I've re-applied the fix.

jacobgkau:
The ghost window issue is fixed now when moving windows to new workspaces below where they are, but not when moving windows to new workspaces above where they are. https://www.dropbox.com/s/63p6b0djuu8wjsu/shell-pr720-screencast2.mkv?dl=0 

Also discovered https://github.com/pop-os/shell/issues/728, which does not appear to be a regression, although the overview change interacts with it a bit (this PR sort of adds a workaround.)

mmstick:
Took a while to figure out why that was happening, but I'm able to drop windows in the overview between workspaces without the ghost windows, while also still being able to drop them on a different display.

jacobgkau (approved):
Regression testing passed. [shell-pr720-run2.txt](https://github.com/pop-os/shell/files/5568748/shell-pr720-run2.txt)

Can't find any more issues with dragging between overviews and workspaces. Stacks on the current workspace are destroyed when dragging to a new in-between workspace (and all stacks on all workspaces are destroyed when dragging to a new first workspace), but those issues are also present in master.

Issue #482, opened 2020-08-05
Windows not moving to new workspace when second display is unplugged


jacobgkau:
**Distribution (run `cat /etc/os-release`):** Pop!_OS 20.04 LTS
 
**Related Application and/or Package Version (run `apt policy pop-shell` or provide the latest commit if building locally):** `0.1.0~1596663403~20.04~c976582` (shell-misc branch) and `0.1.0~1593273798~20.04~9e8ab29` (current stable).

**Issue/Bug Description:** When a second display is unplugged, its tree is supposed to be placed onto a new workspace on the remaining display.

**Steps to reproduce (if you know):**
- Open 8 windows (Firefox, Terminal, Gedit, Pop!_Shop, Files, Contacts, Calendar, Settings).
- Arrange windows in quarter-tiles on both displays.
- Unplug second display.
- The remaining display has windows hanging off of its edges, overlapping, and not able to be moved/resized properly.

**Expected behavior:** The four windows from the second display should move to a new workspace on the remaining display.

**Other Notes:**

mmstick:
Do these monitors have different resolutions?

jacobgkau:
> Do these monitors have different resolutions?

The issue occurs with different resolutions (1080p internal display and 1440p external display), and it still occurs if I make the monitors the same resolution (1080p internal display and 1080p external display). I initially saw the issue on the gaze14 in NVIDIA mode, just confirmed it on the darp6 with Intel graphics.

Today, testing with `990c626`, I'm seeing the four windows from the second display being placed directly on top of the four windows from the primary display (with subsequent issues accessing the windows in the back.)

jacobgkau:
Resolved by https://github.com/pop-os/shell/pull/721.

PR #721, opened 2020-11-13
fix: Improvements to display hotplugging

mmstick:


jacobgkau (approved):
It's a little slow, but workspaces now adjust themselves to be properly sized and visible on the screen when adjusting display positions or plugging/unplugging displays that aren't lined up at the top.

Workspaces on a primary display are now usually unchanged (as they're supposed to be) when plugging in a secondary display. (I thought I saw one case where they started moving to the newly-plugged display instead, but I haven't been able to recreate it.)

This definitely fixes https://github.com/pop-os/shell/issues/482 in the current state. All workspaces from the secondary display are now properly moved to the primary display when the secondary display is unplugged. (Since my secondary display is higher resolution than the primary, I can sometimes get partially-overlapping windows on the primary display depending on how they were arranged on the secondary display and what the windows' minimum sizes are.)

Regression testing passed for Groovy and Focal. (On Groovy, I had to build and install locally to test, since CI didn't build Groovy for whatever reason.) [shell-pr721-groovy.txt](https://github.com/pop-os/shell/files/5593137/shell-pr721-groovy.txt), [shell-pr721-focal.txt](https://github.com/pop-os/shell/files/5593138/shell-pr721-focal.txt)

I did notice one quirk on Focal. If I start with this layout:

- Workspace 1:
  - Display 1: Firefox
  - Display 2: Nautilus
- Workspace 2:
  - Display 1: Terminal
  - Display 2: Pop!_Shop

Screenshot from 2020-11-24 15-54-12

If I navigate to Workspace 2 and then unplug the second display, the second display's workspaces are moved to the primary display, and the workspace with the Pop!_Shop is active:

Screenshot from 2020-11-24 15-54-29

When I plug the second display back in, the windows correctly stay on the primary display, but for some reason, I get an unexpected empty workspace in the middle of my workspaces (with the Pop!_Shop being moved down to a new lower workspace):

Screenshot from 2020-11-24 15-54-58

The result is that I have to manually switch workspaces after plugging the second display back in. (As soon as I switch to one of the populated workspaces, the empty one goes away.)

This doesn't happen on Groovy (Groovy also has the Terminal workspace active, rather than the Pop!_Shop workspace, after unplugging the second display.) I'm going to approve this as-is, since even Focal's behavior is a big improvement over the current version when unplugging displays, but I can re-test if we want to look into that specific case further.

PR #722, opened 2020-11-14
Override terminal launch shortcut to be in line with others

Xanewok:
Before this change, terminal seemed to be the only shortcut not being
overriden (at least on default Ubuntu 20.10) from the Keyboard Shortcuts > Launchers
section and used Ctrl+Alt+T instead of Super+T.
This change aims to make things a bit more consistent and hopefully
facilitate transition from other desktop environments (like Xfce).

Originally opened as #708 but it seems I can't edit the PR to change the base branch, hopefully that's okay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ These are key to many of the shortcuts utilized by tiling window managers. This 
 - `Super` + `f`: Files
 - `Super` + `e`: Email
 - `Super` + `b`: Web Browser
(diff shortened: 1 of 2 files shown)


mmstick (approved):

jacobgkau (approved):

Xanewok:
Thanks for merging it!

PR #723, opened 2020-11-16
fix: Tab disconnect sometimes tries to disconnect undefined tab

mmstick:
I've occasionally run into a scenario where this codepath is hit with an `undefined` tab. This causes the stack to get stuck without an active window.


jacobgkau (approved):
Regression testing passed. [shell-pr723.txt](https://github.com/pop-os/shell/files/5556233/shell-pr723.txt)

ids1024 (approved):

PR #732, opened 2020-11-18
Install schemas system-wide, and install Gnome Control Center keybindings xml files

ids1024:
See commit messages for detailed explanation and justification. This eliminates the need to patch `gnome-control-center` for `pop-shell`. Updates to the keybindings can be made without having interdependencies between pop-shell and `gnome-control-center` packages (after this one).

This should make several things easier and avoid certain problems.

Requires corresponding change to focal/groovy `gnome-control-center` package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ restart-shell:
 
 update-repository:
 	git fetch origin
-	git reset --hard origin/master_focal
+	git reset --hard origin/master
 	git clean -fd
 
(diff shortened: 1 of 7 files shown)


mmstick (approved):

jacobgkau (approved):
All of the moved keyboard shortcuts still show up in GCC. Regression testing passed on Groovy and Focal.

PR #741, opened 2020-11-23
fix: Handle regression in 3.38.2 backports that caused nautilus to resize without moving

mmstick:
Although we can also wait to see if the 3.38.2 release has fixed the problem.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -253,6 +253,7 @@ export class Ext extends Ecs.System<ExtEvent> {
                     const { x, y, width, height } = movement;
 
                     window.meta.move_resize_frame(true, x, y, width, height);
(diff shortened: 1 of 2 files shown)


ids1024 (approved):

jacobgkau (approved):
This fixes the regression with the GNOME Shell and Mutter updates, and does not cause any other regressions. [gnome-debpatches.txt](https://github.com/pop-os/shell/files/5586140/gnome-debpatches.txt)

Linking the GNOME Shell and Mutter updates here for easier navigation:
https://github.com/pop-os/gnome-shell/pull/49
https://github.com/pop-os/mutter/pull/15

mmstick:
Should we merge everything, or wait for 3.38.2?

PR #753, opened 2020-12-02
Improve search order based display name containing pattern

### LINKED ISSUE #108: Launcher search results are sometimes not great
A good example can be seen when searching for "Files", whether Nautilus is open or not. I would expect Nautilus to be the top result, not Audacity.
Screenshot from 2020-03-24 10-30-06

scaryrawr:
Fixes #108 

image

Same with things like Steam or Edge
image

image

### DIFF
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -111,10 +111,15 @@ export class Launcher extends search.Search {
 
             // Sort the list of matched selections
             this.selections.sort((a, b) => {
-                const a_name = a instanceof window.ShellWindow ? a.name(ext) : a[1].name();
-                const b_name = b instanceof window.ShellWindow ? b.name(ext) : b[1].name();
+                const a_name = a instanceof window.ShellWindow ? a.name(ext).toLowerCase() : a[1].name().toLowerCase();
+                const b_name = b instanceof window.ShellWindow ? b.name(ext).toLowerCase() : b[1].name().toLowerCase();
 
-                return a_name.toLowerCase() > b_name.toLowerCase() ? 1 : 0;
+                const pattern_lower = pattern.toLowerCase()
+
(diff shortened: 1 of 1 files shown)


src/dialog_launcher.ts
```diff
@@ -111,10 +111,15 @@ export class Launcher extends search.Search {
...
             this.selections.sort((a, b) => {
-                const a_name = a instanceof window.ShellWindow ? a.name(ext) : a[1].name();
-                const b_name = b instanceof window.ShellWindow ? b.name(ext) : b[1].name();
+                const a_name = a instanceof window.ShellWindow ? a.name(ext).toLowerCase() : a[1].name().toLowerCase();
+                const b_name = b instanceof window.ShellWindow ? b.name(ext).toLowerCase() : b[1].name().toLowerCase();
 
-                return a_name.toLowerCase() > b_name.toLowerCase() ? 1 : 0;
+                const pattern_lower = pattern.toLowerCase()
+
+                const a_includes = a_name.includes(pattern_lower);
```
scaryrawr, line 119: @mmstick would `indexOf` be better here? We could also sort by "how close to start" the position is. I think it's probably not worth doing unless folks still feel the result order needs improving.

mmstick: Ultimately, I've planned to switch to using the levenshtein distance algorithm to compare and weight results.

jacobgkau (approved):
Good improvement to search results. Regression testing passed.
[shell-pr753-groovy.txt](https://github.com/pop-os/shell/files/5644900/shell-pr753-groovy.txt)
[shell-pr753-focal.txt](https://github.com/pop-os/shell/files/5645171/shell-pr753-focal.txt)

Discovered an unrelated issue https://github.com/pop-os/shell/issues/762 while testing (not a regression.)

PR #764, opened 2020-12-05
feat(launcher): Add data dirs indexer

Abdillah:
Support for NixOS (and maybe Guix?). In which the "desktop folder" is located on each app store folder separately then added into `$XDG_DATA_DIRS` environment variables.

### DIFF
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -20,6 +20,7 @@ import type { AppInfo } from 'app_info';
 const { OK } = result;
 
 const HOME_DIR: string = GLib.get_home_dir();
+const DATA_DIRS: string = GLib.get_system_data_dirs();
 
 /// Search paths for finding applications
 const SEARCH_PATHS: Array<[string, string]> = [
@@ -218,6 +219,18 @@ export class Launcher extends search.Search {
     load_desktop_files() {
         lib.bench("load_desktop_files", () => {
             this.desktop_apps.splice(0);
+            for (const _path of DATA_DIRS) {
+                const path = _path.replace(/\/$/, '') + "/applications";
+                for (const result of app_info.load_desktop_entries(path)) {
+                    if (result.kind == OK) {
+                        const value = result.value;
+                        this.desktop_apps.push(['System', value]);
+                    } else {
+                        const why = result.value;
+                        log.warn(why.context(`failed to load desktop app`).format());
+                    }
+                }
+            }
             for (const [where, path] of SEARCH_PATHS) {
                 for (const result of app_info.load_desktop_entries(path)) {
                     if (result.kind == OK) {
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -219,23 +219,26 @@ export class Launcher extends search.Search {
     load_desktop_files() {
         lib.bench("load_desktop_files", () => {
             this.desktop_apps.splice(0);
-            for (const _path of DATA_DIRS) {
-                const path = _path.replace(/\/$/, '') + "/applications";
+            for (const [where, path] of SEARCH_PATHS) {
                 for (const result of app_info.load_desktop_entries(path)) {
                     if (result.kind == OK) {
                         const value = result.value;
-                        this.desktop_apps.push(['System', value]);
+                        this.desktop_apps.push([where, value]);
                     } else {
                         const why = result.value;
                         log.warn(why.context(`failed to load desktop app`).format());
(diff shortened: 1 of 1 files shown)


mmstick (approved):

jacobgkau (changes requested):
After this change, all applications show up twice in the search results:

Screenshot from 2020-12-07 10-52-41

Screenshot from 2020-12-07 10-54-00

Abdillah:
@jacobgkau Is that because the code added a directory twice? Or because the system place a desktop file into multiple directories?

jacobgkau:
> @jacobgkau Is that because the code added a directory twice? Or because the system place a desktop file into multiple directories?

The above screenshots are from Pop!_OS. Using gedit as an example, there were two .desktop files present on the system, one in `/usr/share/applications` and one in `/usr/share/app-install/desktop`. The latter directory is provided by the `app-install-data` package, and even if I remove that package (leaving only one gedit entry on the system in `/usr/share/applications`), I still have duplicate gedit entries.

Looking at your code, you added a section that I see will load `/usr/share/applications`, but you didn't remove any lines (which would mean the code that was already loading `/usr/share/applications` before your change would still be there somewhere), so I would assume it's that the code with your changes added the directory twice. See [lines 26-38](https://github.com/pop-os/shell/blob/058e30e196ef0ead1127f236ce206ca17afb7636/src/dialog_launcher.ts#L25) for where the .desktop directories were specified already, and [line 234](https://github.com/pop-os/shell/blob/058e30e196ef0ead1127f236ce206ca17afb7636/src/dialog_launcher.ts#L234) immediately under your new code for where desktop entries from those directories were already being loaded in.

A solution might be to remove the code that you're replacing, although it looks like `~/.local/share/applications/` is currently hard-coded and is not part of the $XDG_DATA_DIRS environment variable by default in Pop!_OS, so that one might need to stay hard-coded for now. (The only other difference between the hard-coded list and $XDG_DATA_DIRS is that the environment variable contains `/usr/share/pop`, which is not currently hard-coded; that directory doesn't exist on my systems, but I wouldn't see that causing a problem here.)

mmstick:
The list of dirs could be a `Set<string>` so a directory can't be searched twice

Abdillah:
I can reproduce the problem of duplicate entries. 

> The list of dirs could be a Set<string> so a directory can't be searched twice

I tried to filter the paths of duplicate, but still get the same entries. After checking, while my `/usr/share/applications` is empty, this is due to similar desktop files put on different location that all added to XDG_DATA_DIRS. I thought, the problem more of double desktop files, so I take a safer approach by checking the exec for similar command. I've tested in my desktop and it worked fine.

Please evaluate again @jacobgkau, feedback welcome.

jacobgkau:
If I'm understanding correctly, this should now check all of the hardcoded directories first, then check `$XDG_DATA_DIRS` while filtering out any duplicate entries. Anything in `$XDG_DATA_DIRS` that wasn't already found in a hard-coded directory will be marked as a "System" app.

I'm not seeing duplicates, but if I move a .desktop file from `/usr/share/applications` to `/usr/share/pop/applications` (with `/usr/share/pop` being part of my `$XDG_DATA_DIRS` as previously mentioned), it doesn't show up in the search results.

Screenshot from 2020-12-08 14-05-17

After this change, I'm expecting to be able to place `.desktop` files inside of any of the directories listed in `$XDG_DATA_DIRS` plus `/applications`, so it seems like something's still not working quite right here.

jacobgkau (approved):
If I'm understanding correctly, this should now check all of the hardcoded directories first, then check `$XDG_DATA_DIRS` while filtering out any duplicate entries. Anything in `$XDG_DATA_DIRS` that wasn't already found in a hard-coded directory will be marked as a "System" app.

I can confirm items placed in `$XDG_DATA_DIRS` (+`/applications`) that are not one of the hard-coded locations now show up in the launcher. I am no longer getting duplicate entries from directories that are in both the hard-coded list and `$XDG_DATA_DIRS`. I also do not get duplicate entries if I copy a .desktop file into more than one system-wide location. I can still get a separate entry (labeled `Local`) if I copy a .desktop file into `~/.local/share/applications`.

All of the regression tests are passing for the launcher, so if @mmstick is okay with how this is being done, then it looks good to me. (It seems like it might make sense to eventually replace the hard-coded list with `$XDG_DATA_DIRS` and `$XDG_DATA_HOME`, instead of searching a lot of the same places twice and then filtering out the duplicates.)

PR #774, opened 2020-12-11
feat: Add web query plugin

scaryrawr:
- Adds support for launching a web query from launcher

By prefix:
image

Also by if no results found:
image

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -185,8 +185,13 @@
             <summary>Swap window right</summary>
         </key>
 
-        <!-- Workspace Management -->
+        <!-- Launcher Options -->
+        <key type="s" name="search-engine">
+            <default>"https://www.duckduckgo.com/?q="</default>
+            <summary>Search engine query string</summary>
+        </key>
 
+        <!-- Workspace Management -->
         <key type="as" name="pop-workspace-down">
             <default><![CDATA[['<Super><Shift>Down','<Super><Shift>j']]]></default>
             <summary>Move window to the lower workspace</summary>
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -185,13 +185,8 @@
             <summary>Swap window right</summary>
         </key>
 
-        <!-- Launcher Options -->
-        <key type="s" name="search-engine">
-            <default>"https://www.duckduckgo.com/?q="</default>
-            <summary>Search engine query string</summary>
-        </key>
-
         <!-- Workspace Management -->
+
         <key type="as" name="pop-workspace-down">
             <default><![CDATA[['<Super><Shift>Down','<Super><Shift>j']]]></default>
             <summary>Move window to the lower workspace</summary>
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -143,7 +143,7 @@ export class Launcher extends search.Search {
 
             // Fall back to web query if nothing found
             if (this.options.length == 0) {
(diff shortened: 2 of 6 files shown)


scaryrawr:
Latest commit adjusts description and name:
image

schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
```diff
@@ -185,8 +185,13 @@
             <summary>Swap window right</summary>
         </key>
 
-        <!-- Workspace Management -->
+        <!-- Launcher Options -->
+        <key type="s" name="search-engine">
```
mmstick, line 189: I think we should keep plugin settings out of the Pop Shell's settings

src/dialog_launcher.ts
```diff
@@ -141,6 +141,33 @@ export class Launcher extends search.Search {
             // Truncate excess items from the list
             this.options.splice(this.list_max());
 
+            // Fall back to web query if nothing found
```
mmstick, line 144: Maybe we should have a search-not-found condition for plugins?

src/prefs.ts
```diff
@@ -17,8 +17,16 @@ interface AppWidgets {
     smart_gaps: any,
     snap_to_grid: any,
     window_titles: any,
+    search_engine: any,
 }
 
+const SEARCH_ENGINES: string[] = [
```
mmstick, line 23: What if the search engines are based on the first word in the query?

wiki article
bing search
ddg search
google search

src/dialog_launcher.ts
scaryrawr: Hmmm... I wonder what a good name for a setting would be, kind of the "Default Prefix if no results found".

src/prefs.ts
scaryrawr: This makes sense, and a good way to remove it from the settings panel

src/dialog_launcher.ts
scaryrawr: I've removed this for now since it's a bigger behavior change.

jacobgkau:
The Wikipedia option currently doesn't search Wikipedia, it only opens Wikipedia to the specified page. For example, when I type `wiki: test search`, I would expect to go to `https://en.wikipedia.org/w/index.php?search=test+search` for a list of search results, but instead I go to `https://en.wikipedia.org/wiki/Test_search`, which does not exist. (Vising the `?search=` link with the name of an existing page does redirect to the page.) Should the Wikipedia part of this plugin use a search URL instead of the article URL, or is it only meant to be used with existing pages?

For comparison, searching DuckDuckGo with `!w test search` goes to `https://en.wikipedia.org/wiki/Special:Search?search=test%20search&go=Go&ns0=1`. (This is a different URL than what I get when using Wikipedia's own search box, but appears to be the same page.)

scaryrawr:
> The Wikipedia option currently doesn't search Wikipedia, it only opens Wikipedia to the specified page. For example, when I type `wiki: test search`, I would expect to go to `https://en.wikipedia.org/w/index.php?search=test+search` for a list of search results, but instead I go to `https://en.wikipedia.org/wiki/Test_search`, which does not exist. (Vising the `?search=` link with the name of an existing page does redirect to the page.) Should the Wikipedia part of this plugin use a search URL instead of the article URL, or is it only meant to be used with existing pages?
> 
> For comparison, searching DuckDuckGo with `!w test search` goes to `https://en.wikipedia.org/wiki/Special:Search?search=test%20search&go=Go&ns0=1`. (This is a different URL than what I get when using Wikipedia's own search box, but appears to be the same page.)

I've updated to use the `https://en.wikipedia.org/w/index.php?search=` as the base for wiki

jacobgkau (approved):

mmstick:
I will make one change: `:` shouldn't be necessary to identify a web query. `wiki test search` should be fine

PR #777, opened 2020-12-14
fix(plugin-files): Only show hidden directories when a query starts with .

mmstick:
Hide hidden directories and files unless the last component of a search query starts with "."


jacobgkau (approved):
Change works & regression testing passed on Focal and Groovy.

ids1024 (approved):

PR #780, opened 2020-12-16
fix(tiling): Animations clipping if resizing windows too quickly

### LINKED ISSUE #757: Windows overlap (or have incorrect gaps) if resized faster than animations
**(1) Issue/Bug Description:** When resizing windows with the keyboard in window management mode, resizing faster than the resize animations can cause windows to overlap. (Nautilus is an example that overlaps; Firefox is an example that doesn't overlap, but has no gap behind the active hint.)

1. Quarter-tile four windows, with Nautilus in the top left.
2. Enter window management mode on the top-left window.
3. Press `Super+Left` followed by `Super+Up` in quick succession.

**(3) Expected behavior:** Ideally, the actions would either skip the first animation or queue, but even if they don't, the windows should not overlap.

**(5) Gnome Shell version:** `3.38.1-1ubuntu1pop1~1606763654~20.10~01f4376`

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):** Master (`ee943b8`) and except (`e8fd18f`).

**(7) Where was Pop Shell installed from:** Staging repositories.

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):** 1x 1080p.

**(9) Other Installed/Enabled Extensions:** Pop!_OS defaults only.

**(10) Other Notes:** Video: https://www.dropbox.com/s/d9bknoqerhx9ehb/shell-i757-screencast1.mkv?dl=0

mmstick:
Closes #757

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -655,10 +655,6 @@ export class Ext extends Ecs.System<ExtEvent> {
         this.delete_entity(win);
     }
 
-    on_display_move(_from_id: number, _to_id: number) {
-        if (!this.auto_tiler) return;
-    }
-
     /** Triggered when a window has been focused */
     on_focused(win: Window.ShellWindow) {
         this.exit_modes();
--- a/src/forest.ts
+++ b/src/forest.ts
@@ -17,7 +17,7 @@ import type { Ext } from './extension';
 import { Stack } from './stack';
 
 const { Arena } = arena;
-const { Meta } = imports.gi;
+const { GLib, Meta } = imports.gi;
 const { Movement } = movement;
 
 const DOWN = Movement.DOWN;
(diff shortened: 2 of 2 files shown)


jacobgkau (changes requested):
After this change, I'm seeing a lot of failures to move windows in window management mode. Some windows won't move in certain directions. Example: https://www.dropbox.com/s/kyrlklo73xsmgk7/shell-pr780-screencast1.mkv?dl=0

jacobgkau:
The latest change seemed to reduce the clipping, but I was still able to end up in a bad state after doing two quick resizes enough times:

shell-pr780-screenshot1

Also, the latest change seems to slow down resizing and moving of windows. Especially for movement, it may be slower than users desire.

PR #783, opened 2020-12-22
feat: Support Builtin Plugins, Add Scripts Plugin, Optional System76 Power Integration

mmstick:
- Adds support for builtin plugins, which are plugins that are embedded in Pop Shell to be called internally, rather than invoked externally.
    - This is useful for simple plugins written by us that needn't consume an entire process, and to offer quicker results
- Added a new scripts plugin, as the first and only builtin plugin
    - Scripts are stored in these directories:
        - `/usr/lib/pop-shell/scripts/`
        - `$HOME/.local/share/pop-shell/scripts/`
- The session plugin is now a series of scripts for the scripts plugin
- New package for optional integration with system76-power for graphics switching


jacobgkau (approved):
Regression testing passed on Focal and Groovy. Session actions that have been moved to scripts still work. New `system76-power` options work, and are able to be removed by removing the new `pop-shell-plugins-system76-power` package, which installs as part of a regular `apt upgrade`.

My first thought was that `pop-shell-plugin-system76-power` (singular) might make more sense for the package name (similar to `gnome-shell-extension-...` packages), but the current plural name is fine if each of the four scripts are considered a different "plugin."

mmstick:
I'll change the package name and merge

PR #786, opened 2020-12-23
fix(plugin-recent): Improve recent documents results

mmstick:


jacobgkau (approved):
Not seeing a change in the results that are returned (or the order they're placed in-- seems alphabetical), but the trim is working and the URI being in the description field looks better.

For search, it might be worth searching the filename first, and prioritizing those results over results that matched earlier in the path. For example, in this screenshot, I searched `u`:

after-2

The first two results don't have `u` in the filename, only the third one does. I'm assuming they were sorted that way because of the `u` in `Pictures`, which is earlier in the full path.

mmstick:
Followup PR will fix how the results are filtered

PR #793, opened 2020-12-26
Add pinentry-gtk-2 password dialog in default floating windows exceptions

josqu4red:


jacobgkau:
The class `Pinentry-gtk-2` that this PR would currently except doesn't seem to work in Pop (I had to remove `pinentry-gnome3` to force the usage of `pinentry-gtk2`:

Screenshot from 2020-12-29 10-56-10

The window's class seems to just be `Pinentry`:

Screenshot from 2020-12-29 10-56-19

I made sure I was actually looking at the `pinentry-gtk2` dialog by uninstalling that package and seeing the `pinentry-curses` prompt instead. @josqu4red Are you using a different distribution that has the class for that window set differently?

josqu4red:
@jacobgkau Yes I'm using Pop OS (on system76 hardware), and I added this exception using the GUI tool in Pop 20.10.
I just noticed I have 3 pinentry packages installed (curses, gnome3, gtk2), but the main usage I have is with `pass` password manager CLI tool, and it seems to automatically favor pinentry-gtk2.

Example with `config.json` included:

image

ids1024:
It seems reasonable to have an exception here, but in general one shouldn't be using Gtk2 based programs when there are more modern replacements.

`pinentry-gtk2` isn't installed by default. @josqu4red  have you explicitly installed it, or one of the packages that depend on it?

```
> sudo apt rdepends pinentry-gtk2
pinentry-gtk2
Reverse Depends:
  Suggests: pinentry-doc
  Recommends: xubuntu-desktop
  Recommends: xubuntu-core
  Depends: sylpheed
 |Recommends: kwalletcli
 |Recommends: claws-mail-pgpmime
```

josqu4red:
Ok, I just remembered `pinentry-gtk-2` is explicitely set as my GPG `pinentry-program`. Replaced it with `pinentry-gnom3` for now, which is already floating while not in explicit exceptions.
Closing this PR, thanks for this fine project !

jacobgkau:
Ahh, after setting the GPG pinentry-program to `pinentry-gtk-2` in `~/.gnupg/gpg-agent.conf`, I see the `Pinentry-gtk-2` class as shown in your screenshot. It seems like it might depend on how it's launched, explicitly or as a fallback.

My instinct would be to add both `Pinentry-gtk-2` and `Pinentry` to the list, but it also looks like just adding `Pinentry` does work for the `Pinentry-gtk-2` window due to how they're matched.

josqu4red:
@jacobgkau well in my case the issue is closed, but if you want me to update the PR with your suggestion, just let me know.

Issue #778, opened 2020-12-15
Some window-specific floating exceptions not applied on window open


jacobgkau:
**(1) Issue/Bug Description:** When creating floating exceptions for `Current Window Only` for some apps, once the window is closed and re-opened, it is not floating like it should be. Pressing `Super+G` the first time does not set it to floating, but re-tiles it (which seems like the shell knows it was supposed to be floating.) Pressing `Super+G` a second time floats the window.

**(2) Steps to reproduce (if you know):**
1. Open Nautilus and Terminal.
2. Open the Floating Exceptions dialog, apply a `Current Window Only` exception to either window, and close the Floating Exceptions dialog.
3. Close the window the exception was applied to, and re-open that app (which should launch with the same window by default.)
    - The window will be tiled, when it should be floating.
4. Press `Super+G` to float the window.
    - The window will not float, but may change its tiled position.
5. Press `Super+G` a second time to actually float the window.

**(3) Expected behavior:**
1. Once the window has been added to floating exceptions, if the same window is re-opened, it should float automatically.
2. If a window launches as tiled, pressing `Super+G` one time should change it to floating.

**(4) Distribution (run `cat /etc/os-release`):** Pop!_OS 20.04 and 20.10

**(5) Gnome Shell version:** 3.36.4 and 3.38.2

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**
- `1.1.0~1607723907~20.04~d72a2f3` (20.04 `master`)
- `1.1.0~1608046866~20.04~04299a0` (20.04 `except`)
- `1.1.0~1607723907~20.10~d72a2f3` (20.10 `master`)
- `1.1.0~1608046866~20.10~04299a0` (20.10 `except`)

**(7) Where was Pop Shell installed from:** Staging repositories.

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):** Issue occurs with a single 1080p display.

**(9) Other Installed/Enabled Extensions:** Pop!_OS defaults.

**(10) Other Notes:** Video example: https://www.dropbox.com/s/hnxjbvelmdxkggp/shell-pr719-screencast2.webm?dl=0 

When the window fails to float the first time `Super+G` is pressed, it also seems to lose its previous saved position (when it does float the second time `Super+G` is pressed, it stays in the same location and size.)

msdrigg:
[...]

PR #795, opened 2020-12-28
feat(tiling): Recheck tiling exceptions on window title change

mmstick:
Fixes exceptions for windows that change their titles dynamically after initialization

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -850,7 +850,7 @@ export class Ext extends Ecs.System<ExtEvent> {
     on_grab_end(meta: Meta.Window, op?: any) {
         let win = this.get_window(meta);
 
-        if (null === win || !win.is_tilable(this)) {
+        if (null === win || !win.is_tilable(this) || this.contains_tag(win.entity, Tags.Floating)) {
             return;
         }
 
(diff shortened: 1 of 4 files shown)


jacobgkau:
If I understand correctly, this should fix #778. However, I'm getting inconsistent behavior with it:

- With Nautilus, the exception is not applied at window open, but is applied after the title is changed to something else and then back to what I've excepted. https://www.dropbox.com/s/qrwy13b51vjavn2/shell-pr795-screencast1.mkv?dl=0 
- With Termianl, the exception is sometimes applied at window open and sometimes not. When it's not applied, window management mode seems to treat it as if it's floating, then the shell re-tiles it after I try to move it. This would probably mean a fail for regression testing, since I'm unable to move/resize the window in window management mode. https://www.dropbox.com/s/p1wd242u673mo8h/shell-pr795-screencast2.mkv?dl=0

mmstick:
The problem with Terminal and Nautilus is that the title changes whenever you switch a directory, so window-specific tiling doesn't work very well with it.

jacobgkau:
I agree title-specific floating exceptions with those apps would be impractical. I'm more concerned about the inconsistent behavior after the changes.

mmstick:
Yeah, this does break more than it fixes, so I'm closing

Issue #796, opened 2020-12-28
Feat.:Support tab navigation in launcher


mphojele:

mphojele:
Tab key does file completion

PR #797, opened 2020-12-29
feat(launcher): Add Support for tab/backtab Nav

mphojele:
close: #796


jacobgkau (commented):
This currently breaks tab completion in the file browser plugin.

mphojele:
I wasn't aware of that feature, I was like "why the tab/backtab key not doing nothing?". Alright then, that's cool.

PR #798, opened 2020-12-29
fix(tiling): Improve display update behavior

### LINKED ISSUE #801: Occasional Window Shifting Causes Layout Disruption.
At random, it seems like the shell will jitter a window around, like if I'm just using Chrome on its own in a workspace, it kinda shifts around for a moment and then puts itself back in place. This is frustrating, sure, but the bigger problem with this is when it affects workspaces with multiple windows tiled. I'm seeing windows on the left grow in size and shrink windows on the right, sometimes by just a smidge, other times by a MASSIVE amount. This makes the entire extension utterly unusable, as it forces some windows that can't shrink off screen, while totally disrupting the workflow and causing misclicks.

Use Pop!_Shell with tiling enabled. Have some windows tiled next to each other in a workspace and wait. Sometimes it happens when I open a window in a new workspace while there is a set of tiled windows in another workspace, other times it just happens while I'm doing absolutely zero window moving on any workspace whatsoever.

**(3) Expected behavior:**
Windows should stay exactly where they are set. There should be no refitting or reorganizing at random. Especially if a user explicitly corrects this by setting a size via hotkeys.

 ```
NAME="Pop!_OS"
VERSION="20.10"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.10"
VERSION_ID="20.10"
VERSION_CODENAME=groovy
UBUNTU_CODENAME=groovy
LOGO=distributor-logo-pop-os
```

**(5) Gnome Shell version:**
GNOME Shell 3.38.2

```
pop-shell:
  Installed: 1.1.0~1609259439~20.10~c533b20
  Candidate: 1.1.0~1609259439~20.10~c533b20
  Version table:
 *** 1.1.0~1609259439~20.10~c533b20 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu groovy/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu groovy/main i386 Packages
        100 /var/lib/dpkg/status
```

Pop!_OS

3840x2160 @ 2x scaling.

Stock Pop!_OS extensions.

giphy

mmstick:
Working on improving the sluggishness of responses to display updates, and handling changes that should only update workareas instead of moving windows around

Closes #801

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -145,6 +145,8 @@ export class Ext extends Ecs.System<ExtEvent> {
 
     was_locked: boolean = false;
 
+    private workareas_update: null | SignalID = null
+
     /** Record of misc. global objects and their attached signals */
     private signals: Map<GObject.Object, Array<SignalID>> = new Map();
 
@@ -599,7 +601,7 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     monitor_work_area(monitor: number): Rectangle {
-        const meta = display.get_workspace_manager()
+        const meta = wom
             .get_active_workspace()
             .get_work_area_for_monitor(monitor);
 
@@ -1458,7 +1460,7 @@ export class Ext extends Ecs.System<ExtEvent> {
             return true;
         });
 
-        const workspace_manager = display.get_workspace_manager();
+        const workspace_manager = wom;
 
         for (const [, ws] of iter_workspaces(workspace_manager)) {
             let index = ws.index();
@@ -1794,7 +1796,7 @@ export class Ext extends Ecs.System<ExtEvent> {
         }
     }
 
-    update_display_configuration(_workareas_only: boolean) {
+    update_display_configuration(workareas_only: boolean) {
         if (!this.auto_tiler || sessionMode.isLocked) return
 
         if (this.ignore_display_update) {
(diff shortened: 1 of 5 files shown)


jacobgkau:
This is causing some problems with window arrangement after moving windows between workspaces with `Ctrl-Super-Up` and `Ctrl-Super-Down` (empty areas with no windows, windows spanning multiple displays, etc): https://www.dropbox.com/s/a444y50dw6t99sm/shell-pr798-screencast1.mkv?dl=0 

I've also got similarly poor window arrangement (with empty areas) after restarting the shell with windows already open and tiling enabled, which doesn't happen on master.

mmstick:
Found the cause and that issue is fixed

jacobgkau:
I got a lot farther this time, but I eventually ended up with an empty space again after moving windows between several workspaces:
shell-pr798-screenshot1

I also had a separate instance where after unplugging an external display, the windows that were previously on the external display were not correctly positioned on the remaining internal display:
shell-pr798-screenshot2

The former issue is being a little tricky to reproduce, but the latter is happening consistently. Restart the shell, plug in an external display, move a few windows to the external display, and then unplug the external display (the internal display is 1080p and the external is 1440p.)

mmstick:
I think I've found a better solution that fixes these issues

jacobgkau:
If I start with this layout:

Screenshot from 2020-12-30 13-45-56

...and change the primary display to external (right), then the windows on the external display get bumped to a new workspace:

Screenshot from 2020-12-30 13-46-10

I do not have this issue with the same layout on current stable. I did finish the regression test, and this is the only issue I found.

mmstick:
I wasn't able to replicate the same issues, but I did apply some changes that might have fixed the issues seen

jacobgkau:
I'm not seeing that primary issue anymore, at least with that layout. However, I am now seeing the issue after unplugging the external display again:

Screenshot from 2020-12-31 10-13-07

This is on `7ebbd53`. After unplugging the external display, the windows that were previously on the external display do move to a new workspace, but they are arranged improperly (and are not able to be moved properly).

mmstick:
I'm still testing this, but it's in a much better state now regarding changing primary displays and hotplugging.

mmstick:
Migrations to a new display with different dimensions need to rebalance or they become skewed to one side. Fixed.

jacobgkau:
When changing the primary display, all windows are now moving to the newly-primary display, which should not happen. This was happening this morning with `a25db7a` and is still happening with `9432946`.

jacobgkau (approved):
When changing primary screens, there's a chance the fork will shift horizontally: https://www.dropbox.com/s/nzp1gmmeyys9528/shell-pr798-screencast2.mkv?dl=0 

I was able to reproduce something similar on current stable, so that doesn't seem to be a regression. Regression testing passed:
[shell-pr798-groovy.txt](https://github.com/pop-os/shell/files/5767305/shell-pr798-groovy.txt)
[shell-pr798-focal.txt](https://github.com/pop-os/shell/files/5767333/shell-pr798-focal.txt)

Issue #308, opened 2020-05-06
Horizontal Workspaces Support


plagioriginal:
I've noticed that you remove the shortcuts to move the workspace to the Right or Left, and on Ubuntu 20.04 there is an extension called Horizontal Workspaces that uses them. When I activated the pop-shell extension, those stopped to work, probably because it needs those shortcuts. However, it's not a really cool experience for someone who is using the Horizontal Workspaces.

It would be cool if you guys could make if work for people who use the Horizontal Workspaces on Ubuntu.

acerspyro:
Horizontal workspaces are especially useful on laptops with touchpad gestures.

Vertical three-finger motions are quite awkward to use for browsing. They're fine for showing a menu (single-swipe), but the area is just too short for multiple swipes.

alperenkaplan:
I believe this behavior is the result of default values being empty for switch-to-workspace-left and switch-to-workspace-right keys in dconf. Extension changes the layout to horizontal however there is no keybinding for switching. You can navigate to org.gnome.desktop.wm.keybindings and manually set a keybinding for keys mentioned above. You can use dconf-editor for this purpose. I don't use horizontal workspaces extention however I installed it after reading your issue and I can confirm that it works when you add custom keybindings for those keys.

RafalSkolasinski:
Just comment that this would be also relevant for workspace matrix extension.

hoshsadiq:
Seems setting `move-to-workspace-right` works, but `move-to-workspace-left` doesn't. You'll need to disable Cosmic Workspaces extension, but once done I've set my settings as below to mimic Pop Shell like behaviour but for horizontal workspaces. However, Shift+Super+Left does not work, but Shift+Super+Right works. Same with the equivalent for H and L.
```
gsettings set org.gnome.desktop.wm.keybindings move-to-workspace-left '["<Shift><Super>Left","<Shift><Super>H"]'
gsettings set org.gnome.desktop.wm.keybindings move-to-workspace-right '["<Shift><Super>Right","<Shift><Super>L"]'
```
I'm struggling to find a way to debug why `move-to-workspace-left` is the only one that doesn't work.

By the way, for anyone interested, switching to horizontal workspaces works fine if you add change the following shortcuts (as above, `Cosmic Workspaces` extension needs to be disabled).
```
gsettings set org.gnome.desktop.wm.keybindings switch-to-workspace-right '["<Control><Super>Right","<Control><Super>L"]'
[...]

PR #805, opened 2021-01-02
Feat: Support Horizontal Workspaces

mphojele:
close: #308
Notes
 - The code was already flexible enough to support this, therefore, this can be seen as minor tweak
 - Kept the workspace switching and window placing keys to the default in Gnome 3.38
 - Maybe it's code for the future, since Gnome 40 will be using a horizontal layout

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -197,6 +197,16 @@
             <summary>Move window to the upper workspace</summary>
         </key>
 
+        <key type="as" name="pop-workspace-right">
+            <default><![CDATA[['<Control><Shift><Alt>Right']]]></default>
+            <summary>Move window to workspace on the right</summary>
+        </key>
+
+        <key type="as" name="pop-workspace-left">
+            <default><![CDATA[['<Control><Shift><Alt>Left']]]></default>
+            <summary>Move window to workspace on the left</summary>
+        </key>
+
         <key type="as" name="pop-monitor-down">
             <default><![CDATA[['<Super><Shift><Primary>Down','<Super><Shift><Primary>j']]]></default>
             <summary>Move window to the lower monitor</summary>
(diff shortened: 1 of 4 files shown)


mmstick (approved):

jacobgkau:
After installing the Horizontal Workspaces extension and this PR, I am able to move windows left and right with `Ctrl-Shift-Alt-Left` and `Ctrl-Shift-Alt-Right`, but I'm not able to navigate left and right with `Ctrl-Alt-Left` and `Ctrl-Alt-Right`. Those two shortcuts are being added to `/org/gnome/desktop/wm/keybindings`, but for navigating up and down workspaces, the default shortcuts (`Ctrl-Super-Up` and `Ctrl-Super-Down`) don't match what's shown in that dconf directory (`Super-PgUp`/`Ctrl-Alt-Up` and `Super-PgDn`/`Ctrl-Alt-Down`). Do the move left/move right shortcuts need to be added somewhere else?

mphojele:
No. That dconf directory should show the correct keybindings for switch up/down/left/right as per the configurations in scripts/configure.sh. Maybe try removing .confirm_shortcut_change in the directory where you cloned pop-shell (If you didn't) and then install

WatchMkr:
We currently don't use `Alt` for workspace movement. UX is working on shortcut and gesture design with horizontal workspaces in mind.

jacobgkau:
I'm seeing this work on Ubuntu and Pop now, if I install locally using `make local-install`. So this would allow users of other distributions to use Pop Shell alongside extensions that create horizontal workspaces, but it won't (as of right now) make horizontal workspaces work in Pop!_OS by default. Those extensions will continue to not work in Pop until we fix the default shortcuts elsewhere (looks like it would be in pop-session's gsettings override: https://github.com/pop-os/session/blob/master_focal/debian/pop-session.gsettings-override)

> We currently don't use `Alt` for workspace movement. UX is working on shortcut and gesture design with horizontal workspaces in mind.

Just to be clear, does that mean we should hold on this until we have different shortcuts to use, or should we move forward with assigning these shortcuts to the GNOME defaults for compatibility with other extensions?

(On regular Pop!_OS with no non-default extensions or keyboard shortcuts, adding this PR does have a slight side effect of moving a window to a new last workspace when `Ctrl-Shift-Alt-Right` is pressed; `Ctrl-Shift-Alt-Left` doesn't seem to do anything: https://www.dropbox.com/s/cdi16envy8wfptk/shell-pr805.webm?dl=0)

mmstick:
I'd lean towards waiting before we stabilize a new shortcut

PR #809, opened 2021-01-05
feat(search): Sort search results by levenshtein distance

mmstick:


jacobgkau (approved):
This seems to be working. It's very noticeable when searching for `File` (`File History & Trash` is no longer the first result; instead, the first result is `Files`.) `Settings` now returns `Settings` instead of `NVIDIA X Server Settings`. On the other hand, searching `Cal` used to return `Calendar` as the first result, but now returns `Color` (the GNOME Settings panel.)

It seems like a net improvement to results. No regressions in the launcher.

PR #814, opened 2021-01-08
Correct the description for a few keybindings

foo-dogsquared:
The grammarian in me is just bothered with the inconsistency so I corrected it. ;p


jacobgkau (approved):
These look like they were minor typos. Thank you for the correction!

ids1024 (approved):
This is clearly right, and already the capitalization used in https://github.com/pop-os/shell/blob/master/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml

PR #824, opened 2021-01-12
fix(app-specific): Firefox has a rendering issue on some systems that requires some jiggling

### LINKED ISSUE #388: Firefox window not updating in latest pop_shell update
 pop 20.04
 
 
Firefox

**Issue/Bug Description:**
When resuming from suspend, my firefox windows are transparent or getting transparent soon after resuming.

Open FF. Suspend PC. Resume wake

**Expected behavior:**
Normal FF windows

Screenshot-20200524171736-1108x470

**Other Notes:**

mmstick:
While not an issue with Pop Shell directly, Firefox encounters a rendering issue some systems when the window is opened and resized . Unsure on a way to detect when Firefox has misrendered, but this will add a check for Firefox that will maximize and unmaximize the window if it's attached to an empty workspace.

Closes #388

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -181,6 +181,20 @@ export class AutoTiler {
             id = [id[0], 0]
         }
 
+        // NOTE: Firefox has a rendering heisenbug. Jiggle it after tiling.
+        if (win.meta.get_wm_class()?.includes("Firefox")) {
+            const { GLib, Meta } = imports.gi;
+
+            GLib.timeout_add(GLib.PRIORITY_DEFAULT, 500, () => {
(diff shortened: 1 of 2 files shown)


jacobgkau:
After moving Firefox between workspaces, I'm seeing it shrink to smaller than the proper size after getting un-maximized, and then it doesn't seem to tile with other apps (this was on Groovy): https://www.dropbox.com/s/3r1mh2n7zeue7q9/shell-pr824-screencast1.mkv?dl=0 

Were you ever able to reproduce the rendering issue described in https://github.com/pop-os/shell/issues/388? I'm wondering if the rendering issue applies when tilling is disabled, and whether this fix would still solve it, but I'm not currently seeing it on Intel or NVIDIA machines.

mmstick:
I've seen it happen before, but it's been a while. I haven't been able to replicate it.

PR #833, opened 2021-01-15
feat: Change orientation with mouse & show overlay hint on drag

### LINKED ISSUE #829: Feature request: when moving window with mouse, change tiling orientation according to mouse position
I would love to be able to select the tiling orientation and the windows positions when dragging the window I'm currently moving over another window.

For instance, if I drag a window to the upper part of another window, it would put it over the existing window. If I move it on the left part of the other window, it would be put on the left, etc.

It would be very similar to how we move tabs in VS Code, or terminals in Tilix.

mmstick:
When dragging a window, an overlay will be shown indicating where the window will be placed in the tree on drop.
This will make it intuitive to set the tiling orientation on drop of a window, rather than requiring a keyboard

Closes #829

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -6,6 +6,7 @@ import * as log from 'log';
 import * as node from 'node';
 import * as result from 'result';
 import * as stack from 'stack';
+import * as Rect from 'rectangle';
 
 import type { Entity } from 'ecs';
 import type { Ext } from 'extension';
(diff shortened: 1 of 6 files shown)


jackpot51 (approved):

jacobgkau (changes requested):
After maximizing a window with the mouse (dragging to the top bar), I'm seeing movement overlays over the maximized window wherever I move my mouse cursor (when I am not currently dragging anything):

Screenshot from 2021-01-15 11-16-00

Screenshot from 2021-01-15 11-16-12

PR #834, opened 2021-01-15
center the icon to better align with the other panel icons

merrickluo:
before:

after:


jacobgkau (approved):
Icons look identical but are now vertically centered; looks good to me.

Before:
shell-pr834-tiled-1-before
shell-pr834-untiled-1-before

After:
shell-pr834-tiled-2-after
shell-pr834-untiled-2-after

ids1024 (approved):

PR #843, opened 2021-01-21
Fix pointer jumping & title bar minimizing

mmstick:
Fixes two recent regressions

Closes #841
Closes #842

PR #844, opened 2021-01-22
Support stacking and tiling orientation of windows by mouse

mmstick:


jacobgkau:
This seems related to https://github.com/pop-os/shell/pull/843, but if I maximize a window, then click the top bar (and get the mouse cursor moved to the top left of the workspace), then I still see overlays when I'm not dragging anything: https://www.dropbox.com/s/yylqr82wku1xjmb/shell-pr844-screencast1.mkv?dl=0

mmstick:
Overlays shall not be shown anymore

jacobgkau:
Confirmed that fixed the overlay issue.

When dragging a window onto the center of another window above/below (in the same column), the windows swap positions. When dragging into a different column, the center stacks. When dragging above/below from an already-stacked window, the result is also a stack. https://www.dropbox.com/s/yylqr82wku1xjmb/shell-pr844-screencast1.mkv?dl=0 

This seems a little visually inconsistent without a clear indication of whether I'm going to end up with swapped positions or a stack. Can dragging a window up/down in a column show zones for where the window will be placed, the same as when dragging into a different column? (Right now, I can't change orientation of a column with the mouse either, since no zones are displayed when dragging in the same column.)

mmstick:
The only way to make it consistent would be to remove the swap feature on drop by mouse, and have it stack instead of swap. Does that sound like a good idea?

jacobgkau:
> The only way to make it consistent would be to remove the swap feature on drop by mouse, and have it stack instead of swap. Does that sound like a good idea?

If I drag a window from one column to another, I can drag it to the top, bottom, left, or right to place it side-by-side, or the center to stack it:

shell-pr844-gif1

What I'm expecting is to see those same zones for any window. If the windows are in the same column (and I'm starting from the bottom window), then left or right would change orientation (in the appropriate direction), top would swap the windows, and bottom would do nothing (since the window I'm dragging is already below the one I'm dropping it onto.) Does that make sense? That would not leave any operation unable to be performed by the mouse.

PR #865, opened 2021-02-03
Stacking & Orientation Toggling via Mouse

mmstick:


jacobgkau:
This does address the feedback from https://github.com/pop-os/shell/pull/844. It's now possible to either stack or rearrange/swap from any position by dropping onto the appropriate area.

The overlay now seems to reflect the absolute new position of the window being dragged/dropped, rather than the relative position compared to the window it's being dropped onto. (I.e., if we're swapping positions, the overlay seems to cover the window we're dropping onto; if we're stacking, the overlay encompasses the entire area that's about to become a stack.) This is farther away from the Tilix example than the previous iteration was, and I personally found it less intuitive, but it works.

I am seeing windows overlapping each other after a stack in the bottom half of a column is destroyed with the mouse (by dragging the second-to-last window in the stack somewhere else): https://www.dropbox.com/s/qx03qt2geis4v76/shell-pr865-screencast3.mkv?dl=0 

That video also shows that if I try to drag a window above/below or left/right of a stack, the overlay suggests that it's possible, but the window goes into the stack instead.

mmstick:
@jacobgkau If those were the only issues, I think we're done here 🎉

mmstick:
I found an issue

jacobgkau:
Not sure if this is the problem you found, but while moving below/left/right of stacks is working, trying to move above a stack that's at the top of a column results in a ghost window: https://www.dropbox.com/s/autfikgx4db5k2a/shell-pr865-screencast4.mkv?dl=0 

When thinking about the expected behavior, this is a case where the overlay would be covering either the entire stack if we're stacking the window, or the entire stack if we're moving above the stack (swapping positions with the stack.) This is where it would make more sense for the overlay to always cover only the top half of the window we're dropping onto if the dragged window is about to move above that window (even if the windows are already in the same column.)

I'm also getting a ghost window if I move a window to the left of a stack on the left side of the screen, then move the window to the right of the stack (once again, this would ideally "swap positions" as the stack would be pushed to the left edge once the other window's no longer there.)

shell-pr865-screenshot1

mmstick:
Yep, that's the one

mmstick:
Found for the solution for this one. Changing the orientation of tiling when one of the siblings is a stack is functional now.

jacobgkau:
Mouse movement is now working consistently and reliably. I noticed that even when using the keyboard, removing the second-to-last window now destroys the stack (where it previously did not):

stack-create-destroy

I don't think that's an issue at all, but wanted to note it.

Unrelated regression, when moving windows between displays with the mouse, the window always moves to the bottom right of the display, regardless of where it's dropped: https://www.dropbox.com/s/gbqj90bkq59ln1k/shell-pr865-screencast5.mkv?dl=0

mmstick:
Is this good to merge?

jacobgkau:
@mmstick No, see the last video that I linked. When dragging a window between displays, the window goes to the bottom right of the display, not where the overlay suggests it will go: https://www.dropbox.com/s/gbqj90bkq59ln1k/shell-pr865-screencast5.mkv?dl=0 

This may not technically be a regression since there was no overlay to suggest it was possible to choose where the window will end up before, but it doesn't seem like expected behavior now.

mmstick:
Okay, I'll fix that then

jacobgkau:
Looks almost perfect now. Moving between displays with the mouse now follows the overlay, moving between displays in the overview still works as expected. Regression testing passed:
[shell-pr865-groovy.txt](https://github.com/pop-os/shell/files/5952442/shell-pr865-groovy.txt)
[shell-pr865-focal.txt](https://github.com/pop-os/shell/files/5952600/shell-pr865-focal.txt)

I just noticed that when moving between displays with the mouse, if I'm moving to a display with no windows on it, then the overlay doesn't want to acknowledge the empty display:
shell-pr865-gif2

That seems fairly minor, would it be simple to fix here or should I open an issue for it?

mmstick:
Last fix being push soon

jackpot51 (approved):

mmstick:
Found one issue where a window moved to a display that had only one window would not set the orientation of the placement because it isn't a sibling of the window being attached to. That issue is now fixed. Top level forks without a right branch that are being asked to place in a certain orientation will now set that orientation after attachment.

jacobgkau:
With `2b70820`, I am unable to move a window from one display to another with the mouse if it's the only window open on any of the displays.
shell-pr865-gif3

Also, the overlay usually doesn't cover what will become a gap:
mpv-shot0002
But when moving to an empty display, the overlay reaches to the edges of the workspace, as if there won't be a gap:
mpv-shot0005

The second one isn't necessarily a huge deal, the first one seems like an issue.

mmstick:
Fixed both issues

jacobgkau (approved):
Looks great. Not seeing any more issues.
[shell-pr865-groovy-run2.txt](https://github.com/pop-os/shell/files/5954909/shell-pr865-groovy-run2.txt)
[shell-pr865-focal-run2.txt](https://github.com/pop-os/shell/files/5954953/shell-pr865-focal-run2.txt)

PR #867, opened 2021-02-03
feat: Support numpad in default keybindings

ids1024:
Keys on the numpad are considered distinct, but users seem to expect them to work the same way. (See recent discussion on  Mattermost.)

I opened an issue with Mutter suggesting these could be treated equally there: https://gitlab.gnome.org/GNOME/mutter/-/issues/1631

Alternately we could not make a change here, and instead patch Mutter. But I'm included to wait to see what upstream Gnome thinks of that. The only disadvantage I can think of to this change is a bit more clutter in Gnome Control Center when listing bindings.


mmstick (approved):

bflanagin (approved):
Works as expected

PR #871, opened 2021-02-06
Add libraries.io websearch

varac:


jacobgkau (approved):
Works as expected for searching libraries.io, no regressions in the launcher.

PR #874, opened 2021-02-11
fix: Mouse dropping from stack to adjacent sibling

mmstick:
If you have three windows, and two are stacked, moving a window from the stack to attach to the adjacent window will now work instead of swapping positions.

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -456,7 +456,7 @@ export class AutoTiler {
                         return true
                     }
                 }
-            } else if (is_sibling) {
+            } else if (is_sibling && win.stack === null) {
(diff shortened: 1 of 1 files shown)


bflanagin:
I haven't run a full regression test yet, but initial tests look good. Its a little difficult to figure out where to place the window in order to trigger the re-stacking, but the highlight does a good job of informing you of whats about to happen.

bflanagin (approved):
No regressions found. 

For good measure after testing the suggested window configuration I also tested more than three windows and various stacked layouts. All worked as expected after the slight learning curve mentioned in my first comment.

jackpot51 (approved):

PR #882, opened 2021-02-16
fix: Don't show border on maximized windows

coremaze:
Hide the active hint border on maximized windows. 

Applying a border to a maximized window is not helpful unless multiple monitors are being used. Even on multi-monitor setups, applying a border to a maximized window is more distracting than helpful to me, though that may be a matter of opinion.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -779,7 +779,7 @@ export class Ext extends Ecs.System<ExtEvent> {
 
         const focus = this.focus_window();
         if (focus) {
-            focus.show_border();
+            focus.update_border_state();
         }
     }
 
--- a/src/window.ts
+++ b/src/window.ts
@@ -301,7 +301,7 @@ export class ShellWindow {
                 ext.tween_signals.delete(entity_string);
                 if (meta.appears_focused) {
                     this.update_border_layout();
-                    this.show_border();
+                    this.update_border_state();
                 }
             };
 
@@ -390,16 +390,19 @@ export class ShellWindow {
         })
     }
 
-    show_border() {
+    update_border_state() {
         this.restack();
         if (this.ext.settings.active_hint()) {
             let border = this.border;
             if (!this.meta.is_fullscreen() &&
+                !this.is_maximized() &&
                 !this.meta.minimized &&
-                this.same_workspace()) {
-                if (this.meta.appears_focused) {
+                this.same_workspace() &&
+                this.meta.appears_focused) {
                     border.show();
-                }
+            }
+            else {
+                border.hide();
             }
         }
     }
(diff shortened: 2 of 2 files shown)


jacobgkau:
Thank you for the contribution. Maximized windows having an active hint has been expected behavior since https://github.com/pop-os/shell/pull/544. Full-screened windows, by contrast, do not have an active hint. Another way to get a window without an active hint is by enabling the experimental Smart Gaps feature in the extension settings; this removes both the outer gaps and the active hint when a window is the only window on a workspace.

@mmstick @maria-komarova Thoughts on if this change is something we should move forward with? It would be more consistent with the smart gapped and full-screened states, but it would make it more difficult to determine which window has focus in a multi-display setup.

jacobgkau:
Looks like the consensus to have hints on maximized windows was reached in https://github.com/pop-os/shell/issues/219. (That issue also seems to indicate that a smart gapped window should have a hint, but it wasn't implemented in #544 due to a difficulty in implementation, since Smart Gaps is still experimental.) This may be a situation where if a user finds the hint distracting, they should turn it off using the top-right menu.

ChrisMiuchiz:
> It would be more consistent with the smart gapped and full-screened states, but it would make it more difficult to determine which window has focus in a multi-display setup.

As far as I can tell, the same behavior is intended for full-screen, which leads to the same potential for ambiguity if you have multiple programs in full-screen on multiple displays. If it is considered an acceptable tradeoff for full-screen, maybe it should be for maximized windows too, since they are very similar.

> This may be a situation where if a user finds the hint distracting, they should turn it off using the top-right menu or the launcher.

I might be misunderstanding this, but I _do_ find it to be useful, and as soon as I open two or more windows on the same display, I do want the hint to be present. Having to disable and enable it manually whenever I do this would probably be cumbersome.

jmmaranan:
I think also adding the monitor count would be nice to determine the hint on maximized state (so it supports the multi display). And then adding in a keyboard shortcut would also be nice to toggle it (so it isn't cumbersome to reach out for the menu). E.g. <Alt><Shift>x

nathansgithub:
I agree that the hint is distracting and not useful on a full screened window. I think a good compromise would be what @jmmaranan suggested where the active hint would only appear on a full screen window if there are at least 2 monitors on (especially since most users probably only use it on one monitor). I think either that or turning off all hints for full screen windows would be an improvement.

jmmaranan:
> I agree that the hint is distracting and not useful on a full screened window. I think a good compromise would be what @jmmaranan suggested where the active hint would only appear on a full screen window if there are at least 2 monitors on (especially since most users probably only use it on one monitor). I think either that or turning off all hints for full screen windows would be an improvement.

The active hint supports suppressing itself on fullscreen state. And when working on parts of #219, I thought we originally had the max window state with a very thin border (1px or 2px), and 4-5px on regular focus, but finalized as 3px and then forgot about it. I think that's why the maximized state active hint was not removed.

namefailed:
can we either figure out  this or fix smart gaps?

Issue #849, opened 2021-01-23
Launcher does not learn on my system


kahironimashte:
**(1) Issue/Bug Description:**
Launcher does not learn preferred commands

**(2) Steps to reproduce (if you know):**
Start an app multiple times does not reduce the keys necessary to make app default selection

**(3) Expected behavior:**
After multiple starts typing "f" should default select "firefox"
This works is how the Kupfer Launcher works. 

**(4) Distribution (run `cat /etc/os-release`):**
 VERSION="20.04.1 LTS (Focal Fossa)"

**(5) Gnome Shell version:**
3.36.4

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**

**(7) Where was Pop Shell installed from:**
github

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**

**(9) Other Installed/Enabled Extensions:**

**(10) Other Notes:**

NonLogicalDev:
I would second that, coming from Alfred on MacOS. It honestly brings a ton of fluidity if the results and apps/windows (tricky duality in linux) especially are ordered by frecency (recency + frequency).

If possible can someone knowledgeable at point at a few files where to look for how the window/result list is built? Would be a good way to help someone to take a stab at it.

scaryrawr:
> I would second that, coming from Alfred on MacOS. It honestly brings a ton of fluidity if the results and apps/windows (tricky duality in linux) especially are ordered by frecency (recency + frequency).
> 
> If possible can someone knowledgeable point at a few files where to look for how the window/result list is built? Would be a good way to help someone to take a stab at it.

I think you're interested in [dialog_launcher.ts](https://github.com/pop-os/shell/blob/master/src/dialog_launcher.ts#L128) specifically the sorter.

Depending on what else though, there's also a special case when the launcher is opened, but the user hasn't typed anything yet [pattern empty](https://github.com/pop-os/shell/blob/master/src/dialog_launcher.ts#L56), it might be good to have the Window List + frequently launched apps list.

PR #884, opened 2021-02-19
feat: Add most recently used app list

scaryrawr:
- "Fixes" #849 
- Adds a small MRU List to a new config file
  - Currently set at 5 items, may be good to try a different count
- Search results will prefer something from the MRU List
  - If multiple items are in the MRU List (or neither) items sorted by default method

### DIFF
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -147,14 +147,18 @@ export class Launcher extends search.Search {
                 }
 
                 if (this.mru_list) {
-                    const a_recent = this.mru_list.is_recent(a);
-                    const b_recent = this.mru_list.is_recent(b);
-                    if (a_recent && !b_recent) {
+                    const a_recent = this.mru_list.recent_score(a);
+                    const b_recent = this.mru_list.recent_score(b);
+                    if (a_recent !== undefined && b_recent === undefined) {
                         return -1;
                     }
 
-                    if (b_recent && !a_recent) {
+                    if (b_recent !== undefined && a_recent === undefined) {
                         return 1;
+                    }
+
+                    if (a_recent !== undefined && b_recent !== undefined) {
+                        return a_recent > b_recent ? 1 : -1;
                     }
                 }
 
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -44,6 +44,7 @@ export class Launcher extends search.Search {
     service: launch.LauncherService
     last_plugin: null | plugins.Plugin.Source
     mode: number;
+    mru_list?: mru.MruList;
 
     constructor(ext: Ext) {
         let cancel = () => {
@@ -126,7 +127,7 @@ export class Launcher extends search.Search {
                 }
             }
 
-            const mru_list = new mru.MruList();
+            this.mru_list = this.mru_list ?? new mru.MruList();
             const sorter = (a: launch.SearchOption, b: launch.SearchOption) => {
                 const a_name = a.title.toLowerCase()
                 const b_name = b.title.toLowerCase()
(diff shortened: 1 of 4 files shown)


src/mru_app_list.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const id = option.id.app.filename;
+    const index = this.entries.indexOf(id);
+    if (index >= 0) {
+      this.entries.splice(index, 1);
+    }
+
+    this.entries.push(id);
+
+    const over = this.entries.length - 5;
```
scaryrawr, line 29: Should probably move 5 to a constant variable at the top of the file to make it easier to find and adjust the list length :rofl:

src/dialog_launcher.ts
```diff
@@ -143,6 +146,18 @@ export class Launcher extends search.Search {
                     b_name_weight = Math.min(b_name_weight, levenshtein.compare(pattern_lower, b.description.toLowerCase()))
                 }
 
+                if (this.mru_list) {
+                    const a_recent = this.mru_list.is_recent(a);
```
scaryrawr, line 150: Since the underlying list is actually in an order based on most recently used, maybe instead of a "bool" I should return the index so we can rank based on that?

scaryrawr:
I think another way to do this is keep track of app+launch count, so most frequently used vs recently used. I just feel recently used is easier :thinking: and can have a size limit, where frequent count would need a bit more thought process.

jacobgkau (approved):
This only keeps track of the last five apps launched through the launcher (so apps launched via the Activities menu or other methods won't be preferred in the launcher search results.)

With that limitation, this does seem to work as advertised. Apps are added to the MRU list as they're opened. If an app was already in the list, it's moved to the end of the list when opened. Only five apps are in the list at a time, and after the list is full, the oldest app is removed when a new one's added. Apps in the list are sorted at the top of search results (after open windows, but before other launchers not in the list.)

This doesn't cause any regressions with other launcher functions, so I'm approving (this still needs engineering review.)

I could potentially see a `Clear Recent Apps` shortcut in the launcher being useful. (I noticed that when removing the `mru_apps.json` file, if I launch an app without first restarting the extension, a full list of five gets re-populated into the list; actually clearing the list requires removing the file and then restarting GNOME Shell.) I could also imagine some users might want to be able to turn this functionality off, but I don't want to overcomplicate things either.

scaryrawr:
> This only keeps track of the last five apps launched through the launcher (so apps launched via the Activities menu or other methods won't be preferred in the launcher search results.)
> 
> With that limitation, this does seem to work as advertised. Apps are added to the MRU list as they're opened. If an app was already in the list, it's moved to the end of the list when opened. Only five apps are in the list at a time, and after the list is full, the oldest app is removed when a new one's added. Apps in the list are sorted at the top of search results (after open windows, but before other launchers not in the list.)
> 
> This doesn't cause any regressions with other launcher functions, so I'm approving (this still needs engineering review.)
> 
> I could potentially see a `Clear Recent Apps` shortcut in the launcher being useful. (I noticed that when removing the `mru_apps.json` file, if I launch an app without first restarting the extension, a full list of five gets re-populated into the list; actually clearing the list requires removing the file and then restarting GNOME Shell.) I could also imagine some users might want to be able to turn this functionality off, but I don't want to overcomplicate things either.

I was wondering how you feel about the count being 5? I picked 5 originally just to prevent other apps from getting pushed too far down, but for folks who's workflow requires a larger number of apps, 5 might be too few.

I think the `Clear Recent Apps` entry definitely makes sense.

For toggling the functionality off, would it be something to add to the preferences window (It looks like it's only settings for auto tiling at the moment)? I'm curious on if a setting is added, should there be tabs or just different sections with titles in the preferences window?

WatchMkr:
`Clear Recent Apps` has no context for the user. Please exclude. We can consider launcher settings separately.

NonLogicalDev:
I feel like an MRU will not really work for solving #849 in earnest, but it is a good start.

I would first suggest to at least make an alteration of using Frecency over MRU.
If a user repeatedly keeps using the same search result, the system should figure out that that result should rank higher.
If a user starts to select a different result for the same query the system should recalibrate.

Ideal system would be able to infer what user wants based on the query, so that the result after a few selects will never be more that two arrow presses away.

Ideally this should also be done per query / per string prefix.
If a user types `fire` and selects firefox.
It should bump the weight of firefox for all prefixes:
`f`, `fi`, `fir`, and of course `fire` 

(if Frecent cache is kept per prefix then 5-10 results limit is probably fine)

If this system further gets extended to window selections, it can be absolutely amazing.

mmstick:
Closing because the JS launcher is about to be replaced by a Rust launcher: https://github.com/pop-os/launcher/

PR #893, opened 2021-03-01
Mappify web plugin table

scaryrawr:
- Move switch case to be a map instead
- Less chance of forgetting a break :safety_vest: 
- Add npm, github gist, github, and flathub


jacobgkau (approved):
Thank you for the contribution! Existing search providers still work with the new map. The four new search providers also work.

Seems like it might be good to alphabetize the list at some point for easier reading/maintaining.

Issue #856, opened 2021-01-27
Tile Windows Toggle not updating when using hotkeys


bflanagin:
**(1) Issue/Bug Description:**

When using `super+y` the visual toggle (Check button) only updates when switching from titling to floating but not the reverse. 

See screenshot: 

![Screenshot from 2021-01-27 10-12-51](https://user-images.githubusercontent.com/10382037/106019752-9cf43180-6088-11eb-9af0-333c1ab88621.png)

**(2) Steps to reproduce (if you know):**

Turn on Tiling using hotkey and check the visual status (issue appears)
Turn off Tiling using hotkey, turn on tiling using UI 
Turn off Tiling using hotkey, check tiling using UI (it has updated)
Toggle only using hotkeys, checking UI. (no updates occur)

**(3) Expected behavior:**
Visual toggle should be in sync with tiling status.

**(4) Distribution (run `cat /etc/os-release`):**
 
NAME="Pop!_OS"
VERSION="20.10"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.10"
VERSION_ID="20.10"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=groovy
UBUNTU_CODENAME=groovy
LOGO=distributor-logo-pop-os

**(5) Gnome Shell version:**

```
gnome-shell:
  Installed: 3.38.2-1pop1~1610298831~20.10~41092e5
  Candidate: 3.38.2-1pop1~1610298831~20.10~41092e5
  Version table:
     3.38.2-1ubuntu1~20.10.1 500
        500 http://us.archive.ubuntu.com/ubuntu groovy-updates/main amd64 Packages
 *** 3.38.2-1pop1~1610298831~20.10~41092e5 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu groovy/main amd64 Packages
        100 /var/lib/dpkg/status
     3.38.1-1ubuntu1 500
        500 http://us.archive.ubuntu.com/ubuntu groovy/main amd64 Packages
```

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**

```
pop-shell:
  Installed: 1.1.0~1611709512~20.10~673b87a
[...]

PR #910, opened 2021-03-22
Dev Rollup

mmstick:
Closes #712 
Closes #856 
Closes #877 
Closes #889 
Closes #906 

- Also fixes a maximized stacked window causing the tab to remain invisible if alt+tabbed to an sibling window

PR #931, opened 2021-04-08
fix: Update `Tile windows` toggle when changed with key binding

ids1024:
Fixes https://github.com/pop-os/shell/issues/856


bflanagin:
*scratches head* I think this might also be fixed in https://github.com/pop-os/shell/pull/910.

Issue #932, opened 2021-04-09
Unable to resize or move window with mouse on Arch Gnome40.


laichiaheng:
**(1) Issue/Bug Description:**
I can't resize or move the window by dragging it with my mouse.

**(2) Steps to reproduce (if you know):**
Try to resize or move the window with my mouse.

**(3) Expected behavior:**
It resizes or moves the window.

**(4) Distribution (run `cat /etc/os-release`):**
```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://www.archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
```

**(5) Gnome Shell version:**
40.0.0

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**

```
Name            : gnome-shell-extension-pop-shell-git
Version         : r677.163d207-1
Description     : Pop Shell - Tiling window management in Gnome (WIP)
Architecture    : any
URL             : https://github.com/pop-os/shell
Licenses        : GPLv3
Groups          : None
Provides        : None
Depends On      : gnome-shell  pop-shell-shortcuts-git
Optional Deps   : gnome-shell-extensions: native-window extension for improved window
                  placement [installed]
Required By     : None
Optional For    : None
Conflicts With  : gnome-shell-extension-pop-shell
Replaces        : None
Installed Size  : 1049.04 KiB
Packager        : Unknown Packager
Build Date      : Fri Apr 9 16:46:46 2021
Install Date    : Fri Apr 9 16:46:53 2021
Install Reason  : Explicitly installed
Install Script  : Yes
Validated By    : None
```

**(7) Where was Pop Shell installed from:**
AUR

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**
1 x 1080p

**(9) Other Installed/Enabled Extensions:**
cpufreq, freon, gamemode, gsconnect, sound-output-device-chooser, appindicator

**(10) Other Notes:**

laichiaheng:
Can anyone help?

VeryLongNicknameSuchWow:
[...]

PR #933, opened 2021-04-10
WIP: support for Gnome-shell 40.0

hmmanshh:
included gnome-shell 40.0 in metadata

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -6,6 +6,7 @@
     "settings-schema": "org.gnome.shell.extensions.pop-shell",
     "shell-version": [
         "3.36",
-        "40.beta"
+        "40.beta",
+        "40.0"
     ]
 }


HideyoshiNakazone (commented):
Not enough to be considered Support for Gnome-shell 40.0, since the issue #932

jacobgkau (approved):
I agree the commit message is misleading, since this doesn't actually add "support." However, the metadata change looks fine. (I was able to install the extension on Arch with released GNOME 40 without this change, though.)

Looking at other extensions for reference, some have dropped 40.beta as they've added 40, for example Ubuntu's [appindicator extension](https://github.com/ubuntu/gnome-shell-extension-appindicator/blob/7b5afbf038a67a64457bccb96d4dd4a1b8e55ef6/metadata.json).

@mmstick will probably know if we want to format this differently.

mmstick:
Yes we need to support GNOME 40 before we broadcast supporting it

Leleat:
> Looking at other extensions for reference, some have dropped 40.beta as they've added 40, for example Ubuntu's appindicator extension.

It was only necessary for extensions in the beginning of the development of GNOME (Shell) 40 because the version check was 'stricter'. It was changed later to now only look at the major version number. https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1719

PS. `40` is better than `40.0`, because EGO has a 'stricter' check than GNOME Shell... at least for now(?). But since Pop Shell isn't on EGO, it doesn't really matter.

PR #935, opened 2021-04-13
feat: Close launcher on super key press

mmstick:
Permits closing the launcher when pressing the overlay key


jacobgkau (approved):
Working as expected, no regressions in the launcher from this change and other shortcuts are still working normally.

ids1024 (approved):

PR #949, opened 2021-04-20
<Alt>F4 keybinding should be preserved

yasserkaddour:
I think it's better to preserve the familiar and popular <Alt>F4 shortcut. 

Thank for all your work and this extension :1st_place_medal:

### DIFF
--- a/scripts/configure.sh
+++ b/scripts/configure.sh
@@ -95,7 +95,7 @@ set_keybindings() {
     dconf write ${KEYS_MEDIA}/rotate-video-lock-static "@as []"
 
     # Close Window
-    dconf write ${KEYS_GNOME_WM}/close "['<Super>q']"
+    dconf write ${KEYS_GNOME_WM}/close "['<Super>q', '<Alt>F4']"
 }
 
 set_keybindings


bflanagin:
Applied the patch to both 20.04 and 21.04 computer with no effect on either my personal account or a test account.

jacobgkau:
> Applied the patch to both 20.04 and 21.04 with no effect on either my personal account or a test account.

This is only editing `configure.sh`, so it will only apply to non-Pop!_OS users (or anyone cloning and building from the git repo), not the built deb package.

`Alt-F4` is unset in Pop!_OS by default. This script is only supposed to set shortcuts to what they default to on Pop, so I'm not sure if this entirely makes sense (it will activate Alt-F4 on Pop!_OS systems when the `configure.sh` script is run.) Most end users who are using `make local-install`/`configure.sh` are probably on other distributions, but anyone developing against Pop Shell might also be using it on Pop!_OS, so they're going to have a non-default shortcut configured.

Ubuntu uses `Alt-F4` by default, and I can confirm that before this change, that shortcut was disabled, while after the change, it still works. @mmstick Do you think this makes sense to add, or should users who want to use a non-Pop-default shortcut just add it back in their settings after configuring?

mmstick:
I think it's fine to preserve Alt + F4 as an option in addition to the preferred Super + Q

jacobgkau (approved):
I'll just approve this, then, because it's working. We can do a separate PR for pop-session if we decide we want this to be a Pop!_OS default later.

PR #951, opened 2021-04-21
feat: Add support for tab navigation for the launcher

yasserkaddour:
This allow the use of the tab key to navigate between the launcher proposed choices, like in most launcher I know (gnome, rofi, dmenu…). This PR don't conflict with the tab completion of plugins. However, I find the behavior of the file plugin very weird since it complete to path that are not on the top of the list. In terminals and vim tab completion is triggered when it's the only choice possible. 

I would suggest rethinking tab completion, or maybe add a setting to choose the tab key behavior.

Cheers :clap:

### DIFF
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -223,14 +223,16 @@ export class Launcher extends search.Search {
             return false
         };
 
-        let complete = () => {
+        let complete = (): boolean => {
             if (this.last_plugin) {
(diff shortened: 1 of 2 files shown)


mmstick (approved):

jacobgkau (approved):
Allows tab navigation and does not interfere with tab completion. No regressions in the launcher.

PR #1003, opened 2021-06-16
Check keywords when searching apps

jacobgkau:
This is working, but since the sorting algorithm doesn't take keywords into account yet, sometimes results are being sorted lower than they should be.

Screenshot from 2021-06-16 20-37-56

While working on this, I'm being conscious not to cause a regression with `cal` returning the Color control panel before Calculator or Calendar.

### DIFF
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -129,6 +129,10 @@ export class Launcher extends search.Search {
                 
                 if (retain) {
                     const generic = app.generic_name();
+                    let keywords = "";
+                    if app.keywords() {
+                        keywords = app.keywords();
+                    }
 
                     const button = new launch.SearchOption(
                         name,
@@ -136,7 +140,8 @@ export class Launcher extends search.Search {
                         'application-default-symbolic',
                         { gicon: app.icon() },
                         this.icon_size(),
-                        { app }
+                        { app },
+                        keywords
                     )
 
                     DedicatedGPU.addPopup(app, button.widget)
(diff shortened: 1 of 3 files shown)


src/dialog_launcher.ts
```diff
@@ -110,10 +110,23 @@ export class Launcher extends search.Search {
...
+                    for (const keyword of app.keywords()) {
+                        const keyword_match = keyword.toLowerCase()
+                        if ( keyword_match.startsWith(pattern)
+                             || keyword_match.includes(pattern)
+                             || levenshtein.compare(keyword_match, pattern) < 3) {
+                            retain = true
+                        }
+                    }
+                }
+
```
mmstick, line 129: Maybe a cleaner solution would be to concatenate the name and keywords into an array, and then bail on the first condition of being true. Which could add the item to the list without needing a `retain` variable anymore.

```js
const name = app.name();
const keywords = app.keywords();
const app_items = keywords !== null ? keywords.concat(name) : [name];
for (const item of app_items) { }
```

jacobgkau: I basically did what you suggested, the only difference being that I concatenated keywords onto the name, instead of the other way around. Even though the sorting's not being done here, since we're breaking from the loop at the first match, it still seems better to compare the name before the potentially-long list of keywords.

jacobgkau:
This is behaving how we want it to:
Screenshot from 2021-06-17 16-58-54
Screenshot from 2021-06-17 16-59-14
Screenshot from 2021-06-17 16-59-15
Screenshot from 2021-06-17 16-58-59

The sorting got a little more complicated, but some degree of separation is necessary in order to prioritize results properly. The prioritization is:

1. Name starts with search pattern
2. Name contains search pattern _or_ keyword starts with or contains search pattern
3. Levenshtein comparison of search pattern with name or description (whichever's closer)
4. Levenshtein comparison of search pattern with closest keyword

Items 2 and 3 could be broken out further if necessary, but it seems to be working well right now.

I also added a new regression test to cover the search algorithm. I didn't include GIMP because it's not a default app (and would therefore add hassle when testing on a fresh machine), but the Popsicle test should cover the same case.

PR #1005, opened 2021-06-17
Check exec field when searching, remove hard-coded search directories

jacobgkau:
This fixes the remaining case in https://github.com/pop-os/cosmic/issues/69, searching `pavu` should return `pavucontrol` first. (It was being returned second to Document Scanner without the sorting changes.)

New sorting priority:

1. Name starts with search pattern
2. Name contains search pattern or keyword starts with or contains search pattern
3. Command starts with search pattern
4. Levenshtein comparison of search pattern with name, description, or command (whichever's closest)
5. Levenshtein comparison of search pattern with closest keyword

Opening as draft so I can also take a look at some other concerns that were brought up with searching.

### DIFF
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -23,21 +23,6 @@ const { OK } = result;
 const HOME_DIR: string = GLib.get_home_dir();
 const DATA_DIRS: string = GLib.get_system_data_dirs();
 
-/// Search paths for finding applications
-const SEARCH_PATHS: Array<[string, string]> = [
-    // System-wide
-    ["System", "/usr/share/applications/"],
-    ["System (Local)", "/usr/local/share/applications/"],
-    // User-local
-    ["User", HOME_DIR + "/.local/share/applications/"],
-    // System-wide flatpaks
-    ["Flatpak (System)", "/var/lib/flatpak/exports/share/applications/"],
-    // User-local flatpaks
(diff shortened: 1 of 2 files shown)


jacobgkau:
https://github.com/pop-os/shell/pull/764 added `XDG_DATA_DIRS` to the launcher's search paths, but did not remove the previous hard-coded list of search directories. The deduplication code was only applied to `XDG_DATA_DIRS`, not the hard-coded directories. This lead to an issue [reported here](https://www.reddit.com/r/pop_os/comments/o1stj1/please_let_me_use_gnome_shell_searchlauncher_as/) where copying a .desktop file from `/usr/share/applications` to `~/.local/share/applications` would result in duplicate entries (I mentioned this in #764 as well.)

Our hard-coded list is nearly identical to `XDG_DATA_DIRS` on Pop!_OS by default (`XDG_DATA_DIRS` actually contains one directory not contained in the hard-coded list), so rather than duplicating the deduplication code, I think it makes more sense to just use `XDG_DATA_DIRS` and let this be fully user-configurable. (Done with befb3a7.)

jacobgkau:
The other issue reported in the [Reddit post](https://www.reddit.com/r/pop_os/comments/o1stj1/please_let_me_use_gnome_shell_searchlauncher_as/) was internationalization, which is going to take more significant work. I'm marking this as ready for review so these improvements don't wait on internationalization.

(I think I could get the international app names to be used as keywords for searching _relatively_ easily, but if someone's using a particular language on their system, they probably want the results displayed in that language too.)

PR #1014, opened 2021-06-20
fix(active-hint): bug fixes and feature max state hint

### LINKED ISSUE #990: Active hint out of screen bug
**Distribution (run `cat /etc/os-release`):**  20.10 x86_64

**Issue/Bug Description:**
When a window is moved out of screen to the left, the active hint stays on screen.
Screenshot from 2021-06-03 17-29-56

just move any window out of screen to the left

**Expected behavior:**
This doesn't happen if the window is moved to the right side
Screenshot from 2021-06-03 17-30-10

### LINKED ISSUE #1013: Active hint stays visible when minimizing last window
Not using autotiling. Minimizing window with button. When **last** window is minimized, the border stays put.

**(3) Expected behavior:**
Active hint should dissapear.

 NAME="Pop!_OS"
VERSION="20.10"

**(5) Gnome Shell version:**
GNOME 3.38.3

3440*1440 (21:9) 

**(10) Other Notes:**
Seemed to have been introduced in normal update in the middle of June 2021.

jmmaranan:
Fixes #990 
Fixes #1013 
Implements #882 hide hint by checking monitor count if single on max state - there seems to be a consensus to still display hint on multi-monitor setup

### DIFF
--- a/src/mod.d.ts
+++ b/src/mod.d.ts
@@ -153,9 +153,9 @@ declare namespace Meta {
     }
 
     enum MaximizeFlags {
-        HORIZONTAL,
-        VERTICAL,
-        BOTH
+        HORIZONTAL = 1,
+        VERTICAL = 2,
+        BOTH = 3
     }
 
     enum MotionDirection {
@@ -187,6 +187,7 @@ declare namespace Meta {
         delete(timestamp: number): void;
         get_buffer_rect(): Rectangular;
         get_compositor_private(): Clutter.Actor | null;
+        get_display(): Meta.Display | null;
         get_description(): string;
         get_frame_rect(): Rectangular;
         get_maximized(): number;
--- a/src/window.ts
+++ b/src/window.ts
@@ -242,6 +242,7 @@ export class ShellWindow {
         return maximized;
     }
 
+
     /**
      * Window is maximized, 0 gapped or smart gapped
      */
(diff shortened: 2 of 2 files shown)


jacobgkau:
Not able to recreate #1013 (tested on 20.10 where it was reported.) Confirmed this fixes #990. Also confirmed the active hint now disappears on maximized windows when only one display is in use, but remains on maximized windows (but not fullscreened windows) when multiple displays are in use.

I did discover a regression with half-snapped (Ctrl-Alt-Left/Right) windows in non-tiling mode. The active hint is no longer visible on any edge except for the one in the middle of the screen (which isn't as helpful when two windows are open):

Screenshot from 2021-06-21 09-23-28

Compare this to the current behavior, which treats a half-tiled hint more like a maximized hint:

Screenshot from 2021-06-21 09-22-55

jmmaranan:
@jacobgkau - thanks for catching the new issue on snap left/right. I'll look into it later.

jmmaranan:
@jacobgkau - snapping on edges should now be fixed

jmmaranan:
I am seeing some issues when the gap is updated to 0. Will send additional changes in a while

jmmaranan:
@jacobgkau - fixed regression on gaps = 0

jacobgkau:
I'm not sure this is actually an issue, but with the latest updates on this branch, half-snapping no longer works while tiling is enabled. This actually makes sense, but we may want to hide the half-snapping animation while tiling is enabled so it doesn't look like it's supposed to work.

jmmaranan:
@jacobgkau - I found the issue on the snapping while tiled. I think removing the snap when tiled functionality is harder, because the light-blue overlay comes from gnome-shell and not on pop-shell.

jmmaranan:
The snap when tiled, should be ready to test

jacobgkau (approved):
Half-snapping is working as expected again. The hint improvements (maximize and out-of-screen) are still working as expected as well.

Regression testing passed:
[shell-pr1014-hirsute.txt](https://github.com/pop-os/shell/files/6703087/shell-pr1014-hirsute.txt)
[shell-pr1014-focal.txt](https://github.com/pop-os/shell/files/6703288/shell-pr1014-focal.txt)

Issue #1017, opened 2021-06-21
Launcher: doesn't show photogimp or xampp


antoniocbarbosa:
**(1) Issue/Bug Description:** After pop-shell update, if you search for photogimp or xampp, it doesn't show the programs. But if searching in the app menu shows.

The gimp was installed via flatpak and I added the photogimp patch to it, xampp was installed with the command sudo ./"installer_name" in the opt/lampp directory.

[Photogimp](https://github.com/Diolinux/PhotoGIMP)

**(2) Steps to reproduce (if you know):** Have the gimp with the photogimp patch and xampp, then search for photogimp or xampp.

After update:

![Screenshot from 2021-06-21 10-45-55](https://user-images.githubusercontent.com/50782838/122773153-78d6fe00-d27e-11eb-91f0-72d9d481d91e.png)

![Screenshot from 2021-06-21 10-45-59](https://user-images.githubusercontent.com/50782838/122773162-7bd1ee80-d27e-11eb-95fc-027b8c63e223.png)

**(3) Expected behavior:**

Before update:

![Screenshot from 2021-06-21 10-38-39](https://user-images.githubusercontent.com/50782838/122773200-84c2c000-d27e-11eb-99e8-9bd49c4de388.png)

![Screenshot from 2021-06-21 10-38-55](https://user-images.githubusercontent.com/50782838/122773209-87251a00-d27e-11eb-99e7-1294e7314d64.png)

**(4) Distribution (run `cat /etc/os-release`):** Pop!_OS 20.10

**(5) Gnome Shell version:** 3.38.3

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):** ```1.1.0-1624036583-20.10-93634d6```

**(7) Where was Pop Shell installed from:** Pop!_OS PPA

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):** 1x 720p

**(9) Other Installed/Enabled Extensions:**

![Screenshot from 2021-06-21 09-35-15](https://user-images.githubusercontent.com/50782838/122773602-de2aef00-d27e-11eb-9f62-23633dd36041.png)

**(10) Other Notes:** pop-shell version before upgrade `1.1.0~1623101914~20.10~df8b27b`

leviport:
I believe this may be a duplicate of https://github.com/pop-os/shell/issues/997

A fix for that was made and is still undergoing regression testing. It should be available pretty soon.

[...]

PR #1018, opened 2021-06-21
Check XDG_DATA_HOME in addition to XDG_DATA_DIRS, label results by location

### LINKED ISSUE #1008: Launcher search doesn't show web apps
**How did you upgrade to 21.04? (Fresh install / Upgrade)**

 Upgrade

 
N/A

**Issue/Bug Description:**
I use Google Chrome with several PWAs installed, and when I search for them in the Launcher, they aren't showing up. However, when I search for them in the application menu, they show up.

1. Press the super key.
2. Search for the Spotify web app.
3. The app doesn't show up.

**Expected behavior:**
Show the web app with all of the other suggested apps.

**Other Notes:**
This only recently started happening. However, I'm pretty sure searching for web apps from the Launcher has worked previously.

jacobgkau:
Fixes #1008, #1017, https://github.com/pop-os/beta/issues/183.

Check XDG_DATA_HOME for user-specific `.desktop` files. The `GLib.get_user_data_dir()` function returns `.local/share` by default on Pop!_OS, which follows the [XDG specification](https://specifications.freedesktop.org/basedir-spec/0.6/ar01s03.html). Check DATA_HOME before DATA_DIRS so user-added `.desktop` files take precedence over system-wide ones.

This also restores functionality lost with https://github.com/pop-os/shell/pull/1005; while it makes sense for search directories to be fully configurable, Flatpak apps were no longer being labeled as Flatpaks because the DATA_DIRS function was labeling all apps as `System`. This compares the path of an app with some well-known paths so they can still be labeled appropriately.

### DIFF
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -21,7 +21,8 @@ const { Clutter, Gio, GLib, Meta } = imports.gi
 const { OK } = result;
 
 const HOME_DIR: string = GLib.get_home_dir();
-const DATA_DIRS: string = GLib.get_system_data_dirs();
+const DATA_DIRS_SYSTEM: string = GLib.get_system_data_dirs();
+const DATA_DIRS_USER: string = GLib.get_user_data_dir();
 
 export class Launcher extends search.Search {
     options: Array<launch.SearchOption>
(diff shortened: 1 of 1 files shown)


mmstick (approved):

bflanagin (approved):
Working as expected.

PR #1027, opened 2021-06-24
Dev Rollup

mmstick:
- Remove Slack from exceptions
- Fix TypeScript compilation errors with the launcher
- Improve responsiveness of searches by delaying a search for 300ms after the last input
- Fixing files not appearing in path-based searches due to the query getting lower-cased
- Fixed searches for paths that match a directory only showing the directory and not its contents

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -35,7 +35,6 @@ export const DEFAULT_RULES: Array<FloatRule> = [
     { class: "KotatogramDesktop", title: "Media viewer" },
     { class: "Steam", title: "^((?!Steam).)*$" },
     { class: "TelegramDesktop", title: "Media viewer" },
-    { class: "Slack", title: "Slack | mini panel" },
     { class: "Solaar", },
     { class: "system76-driver", },
     { class: "zoom", },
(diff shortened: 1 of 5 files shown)


jacobgkau:
While there are some input responsiveness issues on `master`, the 300ms delay makes searching seem much slower. Results don't start to show up until I have the entire query typed in, even when I'm typing slower than usual.

https://user-images.githubusercontent.com/7199422/123297873-33603e00-d4d5-11eb-953e-8181ecdcc123.mp4

We might want to consider a smaller delay. 100ms feels much better than 300ms and doesn't seem to have as much input lag as `master` (although the systems I'm testing on don't always much lag to begin with.)

jacobgkau (approved):
The delay isn't ideal, but it is an improvement, especially on systems where input lag was occurring consistently before.

Confirmed this fixes file browsing in directories with capital letters.

Regression testing passed:
[shell-pr1027-hirsute.txt](https://github.com/pop-os/shell/files/6711386/shell-pr1027-hirsute.txt)
[shell-pr1027-focal.txt](https://github.com/pop-os/shell/files/6711387/shell-pr1027-focal.txt)

PR #1031, opened 2021-06-25
Fix search activation delay / Change shortcut label in 21.04

### LINKED ISSUE #983: Pop Shell > Tiling Menu > If Pop >= 21.04 change Launcher shortcut to "Super"

### LINKED ISSUE #1030: Pressing Enter before results load prevents results from loading
**(1) Issue/Bug Description:** Since the 100ms delay for loading search results was added in https://github.com/pop-os/shell/pull/1027, if a search term is typed and `Enter` is pressed before the 100ms delay has passed, the results never load.

1. Open the launcher
2. Type a search term
3. Immediately press Enter/Return, before the results have loaded

**(3) Expected behavior:** Results should load, starting 100ms after either the last letter or Enter is typed (the former seems better.)

**(5) Gnome Shell version:** 3.38

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):** d59e373

**(7) Where was Pop Shell installed from:** Pop!_OS PPA

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):** 1x 1080p

**(9) Other Installed/Enabled Extensions:** COSMIC/Pop defaults only.

**(10) Other Notes:** Tried to grab a screen recording, but this is difficult to recreate when any window is open since the open windows provide default search results, which remain until the actual search results have loaded.

mmstick:
Closes #983 
Closes #1030

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -41,6 +41,7 @@ export const DEFAULT_RULES: Array<FloatRule> = [
     { class: "Gnome-terminal", title: "Preferences – General" },
     { class: "Gjs", title: "Settings" },
     { class: "Io.elementary.sideload", },
+    { class: "Gnome-initial-setup" },
 ];
 
 export interface FloatRule {
--- a/src/dialog_launcher.ts
+++ b/src/dialog_launcher.ts
@@ -77,54 +77,56 @@ export class Launcher extends search.Search {
 
             const pattern = pat.toLowerCase()
 
-            const needles = pattern.split(' ');
+            if (pat !== '?') {
+                const needles = pattern.split(' ');
 
-            const contains_pattern = (haystack: string, needles: Array<string>): boolean => {
-                const hay = haystack.toLowerCase();
-                return needles.every((n) => hay.includes(n));
-            };
+                const contains_pattern = (haystack: string, needles: Array<string>): boolean => {
+                    const hay = haystack.toLowerCase();
+                    return needles.every((n) => hay.includes(n));
+                };
 
-            // Filter matching windows
-            for (const window of ext.tab_list(Meta.TabList.NORMAL, null)) {
-                const retain = contains_pattern(window.name(ext), needles)
-                    || contains_pattern(window.meta.get_title(), needles);
+                // Filter matching windows
+                for (const window of ext.tab_list(Meta.TabList.NORMAL, null)) {
+                    const retain = contains_pattern(window.name(ext), needles)
+                        || contains_pattern(window.meta.get_title(), needles);
 
-                if (retain) {
-                    windows.push(window_selection(ext, window, this.icon_size()))
+                    if (retain) {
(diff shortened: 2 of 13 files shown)


jackpot51 (approved):

jacobgkau (approved):
Confirmed this fixes #983 and #1030, launcher regression testing passed:
[shell-pr1031-launcher-hirsute.txt](https://github.com/pop-os/shell/files/6716565/shell-pr1031-launcher-hirsute.txt)
[shell-pr1031-launcher-focal.txt](https://github.com/pop-os/shell/files/6716566/shell-pr1031-launcher-focal.txt)

It's worth noting that the pop-shell-shortcuts dialog window still shows `Super` + `/` as the Launcher shortcut in the two places it's listed; we may want to incorporate the change there in addition to the main pop-shell menu.

jacobgkau (changes requested):
Actually, wait. #983 says:

```
With the Super key now being customizable, we can either leave it Super + / or use Super when it's set to Super in Pop 21.04 and later and Super + / when it's not.
```

This change is only checking OS version, incorrectly showing `Super` as the shortcut in the drop-down when that key has been changed to open Workspaces or Applications instead.

jacobgkau:
After I change the setting for the Super key, a GNOME Shell restart is required before the top-right menu's label updates.

Also, the help plugin seems like it could be a little more intuitive. This is what's displayed right now:
Screenshot from 2021-06-28 11-09-59
Clicking or pressing Enter on any of the options just clears the search box, which means a user can't go directly from here to using any of the examples shown. Also, the examples are shown as regular expressions that the Launcher checks for, not what actually needs to be typed. Would it be better if selecting a result input one of the possible prefixes into the search box so the user can see how they're supposed to be used?

jackpot51:
I think help should be an array of manually written items for each plugin, so you can click on any and the prefix populates

jacobgkau:
The new prefills from the help items are working fine. The Calculator and Files plugins display immediate feedback when their prefixes are typed, while the other two plugins don't display feedback until more is typed after the prefill, but that's more to do with the plugins than the new help items.

The last thing I'm noticing is the new initial state text:

Screenshot from 2021-06-28 12-27-51

- The first sentence has a period at the end, so why doesn't the second sentence?
- The word `Press` for the question mark sort of implies the user can just push that key, which won't work on a regular QWERTY layout (the Shift key is required to actually input that character.) What you're actually asking them to do is _type_ a question mark.

Would this work instead? `Type to search apps, or type '?' for more options.`

mmstick:
That'd be fine

jacobgkau (approved):
The delay until next shell restart for updating the top-right menu's keyboard shortcut label is something that could be improved in the future, but for now, this is working well.

PR #1068, opened 2021-07-06
adding a keyword to the logout script

d4v3y0rk:
having `log` and `out` in the keywords section is great but if you type the full word logout this script does not show up. I think it should. I *think* this change will fix that.


jacobgkau (approved):
Thank you for the contribution. This makes the logout script show up when searching `logout`, and no other files are being changed.

PR #1077, opened 2021-07-12
Updated Readme

ruhend:
Fixed a number of typos and grammar issues in the readme file.
Removed a few redundant words to improve readability 🔥 

```
 Total number of changes made: 23 instances
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Pop Shell
 
-Pop Shell is a keyboard-driven layer for GNOME Shell which allows for quick and sensible navigation and management of windows. The core feature of Pop Shell is the addition of advanced tiling window management — a feature that has been highly-sought within our community. For many — ourselves included — i3wm has become the leading competitor to the GNOME desktop.
+Pop Shell is a keyboard-driven layer for GNOME Shell which allows for quick and sensible navigation and management of windows. The core feature of Pop Shell is the addition of advanced tiling window management — a feature that has been highly sought within our community. For many — ourselves included — i3wm has become the leading competitor to the GNOME desktop.
 
 Tiling window management in GNOME is virtually nonexistent, which makes the desktop awkward to interact with when your needs exceed that of two windows at a given time. Luckily, GNOME Shell is an extensible desktop with the foundations that make it possible to implement a tiling window manager on top of the desktop.
 
-Therefore, we see an opportunity here to advance the usability of the GNOME desktop to better accomodate the needs of our community with Pop Shell. Advanced tiling window management is a must for the desktop, so we've merged i3-like tiling window management with the GNOME desktop for the best of both worlds.
+Therefore, we see an opportunity here to advance the usability of the GNOME desktop to better accommodate the needs of our community with Pop Shell. Advanced tiling window management is a must for the desktop, so we've merged i3-like tiling window management with the GNOME desktop for the best of both worlds.
 
 [![](./screenshot.webp)](https://raw.githubusercontent.com/pop-os/shell/master/screenshot.webp)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -114,7 +114,7 @@ These are key to many of the shortcuts utilized by tiling window managers. This
 
 > This mode is activated with `Super` + `Return`.
 
-Window management mode activates additional keyboard control over the size and location of the currently-focused window. The behavior of this mode changes slightly based on whether you are in auto-tile mode, or in the default stacking mode. In the default mode, an overlay is displayed snapped to a grid, which represents a possible future location and size of your focused window. This behavior changes slightly in auto-tiling mode, where resizes are performed immediately, and overlays are only shown when swapping windows.
+Window management mode activates additional keyboard control over the size and location of the currently focused window. The behavior of this mode changes slightly based on whether you are in the auto-tile mode or the default stacking mode. In the default mode, an overlay is displayed snapped to a grid, which represents a possible future location and size of your focused window. This behavior changes slightly in the auto-tiling mode where resizes are performed immediately, and overlays are only shown when swapping windows.
```
jacobgkau, line 117: The hyphen between `currently` and `focused` is not grammatically necessary (`ly` already conveys that the next word is also a modifier), so that change is fine.

`the auto-tile mode` doesn't sound quite right to me. `auto-tile mode` is the name of a specific mode, so even though it's not a proper noun, it may not need an article, similar to the non-proper names of specific sports, academic subjects, etc. The reason `the` is used in `the default stacking mode` is because `default` is a definite noun but isn't a specific name; it wouldn't be needed if the word `default` wasn't there. (By the way, we might want to rename "stacking mode" at some point since "stacking" is now a different feature within tiling mode.)

The comma removal in the last sentence is incorrect. We're not referring to `the auto-tiling mode where resizes are performed immediately` as opposed to some other auto-tiling mode where that's not the case; `This behavior changes slightly in auto-tiling mode` is an independent clause, and the rest of the sentence is a dependent clause that elaborates on the change. If you want to remove a comma here, the second one (between `immediately` and `and`) _is_ grammatically incorrect and should be removed (while `overlays are only shown when swapping windows` might sound like an independent clause, the `where` applies to it as well since it's talking about auto-tiling mode.)

README.md
```diff
@@ -138,11 +138,11 @@ Activating this enables the following behaviors:
...
 
 ### Launcher
 
-The launcher is summoned with `Super` + `/`. The search list displays matching windows based on their window name and title, and applications on the system which can be launched. The arrow keys are used to select an application or window from the search list. If it is a window, the selected window will be visually highlighted with an overlay, in the event that two windows have the same name and title. Pressing `Return` on a window will bring that window to focus, switching to its workspace, and unminimizing it if it was minimized.
+The launcher is summoned with `Super` + `/`. The search list displays matching windows based on their window name and title, and applications on the system which can be launched. The arrow keys are used to select an application or window from the search list. If it is a window, the selected window will be visually highlighted with an overlay, if two windows have the same name and title. Pressing `Return` on a window will bring that window to focus, switching to its workspace, and unminimizing it if it was minimized.
```
jacobgkau, line 145: The usage of `in the event that` wasn't gramatically incorrect before, but the change to `if` is fine grammatically. However, that clause is dependent, so the comma before it should be removed.

Furthermore, the actual sentence was incorrect before, and still is: the Launcher will always highlight the selected window, regardless of whether it's unique. `if` could be changed to `in case` to clarify that we're just referring to the anticipated use case of the feature, but I'd lean toward removing that clause (end the sentence after `overlay`.)

README.md
```diff
@@ -210,15 +210,15 @@ There is file `$XDG_CONFIG_HOME/pop-shell/config.json` where you can add the fol
   title: "<Optional Window Title>"
 }
 ```
-For example, doing `xprop` on GNOME Settings (or GNOME Control Center), the WM_CLASS values are `gnome-control-center` and `Gnome-control-center`. Use the second value (Gnome-control-center), which pop-shell will read. `title` is optional.
+For example, doing `xprop` on GNOME Settings (or GNOME Control Center), the WM_CLASS values are `gnome-control-center` and `Gnome-control-center`. Use the second value (Gnome-control-center), which pop-shell will read. The `title` is optional.
```
jacobgkau, line 213: Like `auto-tile mode` earlier, `title` is the name of a specific field here, so `The` is not necessary. Either say ``The `title` field is optional.`` so `The` can operate on `field`, or don't make this change.

jacobgkau (changes requested):
Thank you for the contribution. Grammar is often a second thought in technical projects, so I appreciate the attention you've given here.

I've left feedback on changes that I disagree with, along with an explanation for each. I agree with any changes I haven't left feedback on.

README.md
```diff
@@ -20,14 +20,14 @@ Therefore, we see an opportunity here to advance the usability of the GNOME desk
...
-- [Developers](#developers): Guide for getting started with development
+- [Developers](#developers): Guide for getting started with the development
 ---
 
 ## The Proposal
 
 A proposal for integration of the tiling window management features from Pop Shell into GNOME is currently under development. It will be created as a GitLab issue on GNOME Shell for future discussion, once we have invested our time into producing a functioning prototype, and learned what does and does not work in practice.
 
-Ideally, the features explored in Pop Shell will be available for any environment using Mutter — far extending the half-monitor tiling capability currently present. By starting out as a shell extension, anyone using GNOME Shell can install this onto their system, without having to install a Pop-specific fork of GNOME on their system.
+Ideally, the features explored in Pop Shell will be available for any environment using Mutter — far extending the half-monitor tiling capability currently present. By starting as a shell extension, anyone using GNOME Shell can install this onto their system, without having to install a Pop-specific fork of GNOME on their system.
```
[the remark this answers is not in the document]
ruhend: Yes, got the point. 👍🏼

README.md
ruhend: Alright! 😄

jacobgkau (approved):

Issue #1072, opened 2021-07-09
Pop Shell breaks Activities on Pop!_OS and Ubuntu 20.04 LTS


d0tsaint:
**(1) Issue/Bug Description:**
Shell seems to "break" Ubuntu Activities. When pop shell is enabled and I go to Activities overview by pressing Super or Clicking the top left, I can't exit out again program windows disappear. I haven't had this issue before, it happened yesterday with a fresh Ubuntu install. The only way to exit back to the desktop; is by logging out and in again or using `ALT+F2` and run `r` to restart.  

**(2) Steps to reproduce (if you know):**
Enable pop shell and go to Activities overview. 

**(3) Expected behavior:**
Cannot exit Activities and cannot see Workspaces. 

**(4) Distribution (run `cat /etc/os-release`):**
NAME="Ubuntu"
VERSION="20.04.2 LTS (Focal Fossa)" 

**(5) Gnome Shell version:**
3.36.8

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**
 

**(7) Where was Pop Shell installed from:**
Git clone. 

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**
Single laptop monitor (Lenovo T490s)

**(9) Other Installed/Enabled Extensions:**
None.

dotsainzed:
I can now confirm that this is not an issue on Ubuntu 21.04.

AxcellF:
Not sure if this is helpful, but anyways:

I've experienced the same issue on a XPS 9560 and a VM (both running Ubuntu 20.04.2 LTS). Made sure to remove all other gnome extensions which didn't make any difference.

jixbo:
I am also affected by this issue since I updated to the latest commit, to fix the VLC full screen issue #942.
I'm using Ubuntu 21.04 and wayland, and everything was working fine before with the older shell extension. What commit are you on @d0tsaint ?
I'm on the latest from master, f4894c801900f55f3fc4a8bccb36c19fae188a8c.

jixbo:
Here are the logs of the issue:
```
gnome-shell[54799]: JS ERROR: Exception in callback for signal: trigger: too much recursion
                                             _show_skip_taskbar_windows/Workspace.prototype._isOverviewWindow@/home/jixbo/.local/share/gnome-shell/extensions/pop-shell@system76.com/extension.js:2003:13
[...]

PR #1080, opened 2021-07-12
fix: Only set `default_*` globals if it not already set

ids1024:
Should prevent "too much recursion" errors.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2577,7 +2577,8 @@ function _show_skip_taskbar_windows(ext: Ext) {
         // So it has to be skipped being overriden for now.
 
         // Handle the overview
-        default_isoverviewwindow_ws = Workspace.prototype._isOverviewWindow;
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

jacobgkau:
I'm not sure how to recreate the "too much recursion" errors, but regression testing passed on Pop!_OS 21.04:
[shell-pr1080-hirsute.txt](https://github.com/pop-os/shell/files/6810653/shell-pr1080-hirsute.txt)

Pop!_OS 20.04 LTS is currently experiencing [issues with the Activities menu](https://github.com/pop-os/shell/issues/1072) that are caused by something other than the recursion errors, and this does not fix that issue.

jacobgkau (approved):
Approving since regression testing passed and this reportedly fixes the issue on machines where the recursion error was happening.

PR #1085, opened 2021-07-14
fix: Correct `_isOverviewWindow` on Gnome Shell 3.36

ids1024:
It seems in 3.38, the argument to this function was changed from the `MetaWindowActor` to the `MetaWindow`. This adds a test to correctly get the `MetaWindow` on both 3.36 and 3.38.

This should fix https://github.com/pop-os/shell/issues/1072.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2576,14 +2576,18 @@ function _show_skip_taskbar_windows(ext: Ext) {
         // So it has to be skipped being overriden for now.
 
         // Handle the overview
-	if (!default_isoverviewwindow_ws)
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

jacobgkau (approved):
Regression testing passed on Hirsute: 
[shell-pr1085-hirsute.txt](https://github.com/pop-os/shell/files/6823547/shell-pr1085-groovy.txt)

Sleek set to minimize to tray still displays normally in the Workspaces overview on Hirsute.

Fixes the critical Activities bug on Focal.

PR #1099, opened 2021-07-20
adds Emojipedia and Giphy search

ahoneybun:
This PR adds support for Emojipedia and Giphy search for the Pop Launcher. It has been built and tested on Pop 21.04 with the latest pop-shell build: 1.1.0~1626357201~21.04~c1f2ea4


bflanagin:
Per Levi's point/concern/suggestion yesterday. It looks like there are rating options that can be added to searches [Giphy search](https://developers.giphy.com/docs/api/endpoint#search.) If these are to be added, it might be a good idea to add rating=g to the search.

ahoneybun:
It looks like this would use the API while currently it just uses the [main site](https://giphy.com/search/term).

bflanagin:
Boo, oh well it was worth a shot.

ahoneybun:
I can replace or remove Giphy if it's a big concern.

leviport:
Are we certain we want to add emoji search? Doesn't Gnome already have emoji search/insert built in?

ahoneybun:
I think it may be better to make this a community package instead of in the main code itself as @bflanagin suggested.

PR #1111, opened 2021-07-26
fix: Focus handling improvements to fix inconsistencies in window add / remove

mmstick:
Sometimes when a new window is opened, the window is placed in a prior-focused stack or window, rather than the actively-focused stack or window. Additionally, when a window in a stack is destroyed, sometimes the focus shifts to a completely different window or stack. These changes should fix the focus handling to correctly append to the actively-focused window, and activate a sibling in the actively-focused stack.

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -655,11 +655,13 @@ export class AutoTiler {
             return Err('ignoring focus');
         }
 
-        if (!ext.prev_focused) {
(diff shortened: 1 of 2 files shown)


bflanagin (approved):
Pop-shell regression tests have passed, on 21.04 and 20.04. The method I used to reproduce the issue before applying the fix no longer triggers the issue once the fix is applied.

PR #1121, opened 2021-08-02
improv: Use `imports.misc.config.PACKAGE_VERSION`

ids1024:
Since this exists, it seems better than spawning a subprocess.

This should have no impact on behavior (except theoretically being faster).


jackpot51 (approved):

bflanagin (approved):
No regressions found on either 20.04 or 21.04. 

[shell-regression-checklist-1121-focal.txt](https://github.com/pop-os/shell/files/6926327/shell-regression-checklist-1121-focal.txt)
[shell-regression-checklist-1121-hirsute.txt](https://github.com/pop-os/shell/files/6926328/shell-regression-checklist-1121-hirsute.txt)

Issue #1123, opened 2021-08-03
20.04 Activies crash with Proton games


SavexG:
Hello 
Im on popOS! 20.04 LTS 

https://user-images.githubusercontent.com/65229679/128024420-80710ad0-9f29-416f-8404-78de6e7a800f.mp4

please watch the video the first klick you hear is me pressing the windows key 

Ive been having this issue with the windows key function where when you press it you see all the windows you have opened at once
for me this is very convenient and Ive been using it a lot but since some time it bugs with Fullscreen applications
I used to be able to conveniently switch between Fullscreen windows as well
but now when I press the windows key while having a Fullscreen application running all the windows I have opened (not Fullscreen windows as well) disappear and I'm stuck on that screen as you can see in the video and no matter what I press nothing comes up the only thing I can do then is to restart or press ctrl + alt + f4 and use the terminal but that's not much of use for me 

but anyway restarting is the only fix Ive found to this so far 
idk is this a known bug? 

Im a linux noob if you want me to do anything specific so you can further investigate this issue don't hesitate to contact me

jacobgkau:
This is not happening with Firefox in fullscreen mode, so it's not all fullscreened apps. Geometry Dash is a paid game, and I don't own it to test with.

If you change the `Show Minimize to Tray Windows` option in Extensions (green icon) -> Pop Shell, does that affect the behavior at all?

If you open a terminal (Super + T), type `xprop`, then click on the fullscreen Geometry Dash window, what output do you get? (The `WM_CLASS` and `WM_NAME` lines may be useful if an exception for that app needs to be added.)

Can you please also get logs using these steps?

1. Within Extensions -> Pop Shell, change `Log Level` from `OFF` to `DEBUG`.
2. Restart GNOME Shell with Ctrl-Alt-F2, r, Enter.
3. In a terminal, run `sudo journalctl -f` and enter your password.
4. Recreate the issue.
5. After Activities freezes, restart GNOME Shell with Ctrl-Alt-F2, r, Enter again, and copy/paste the entire terminal output into a text file, then attach that text file to a message.
6. Change `Log Level` back to `OFF` in the settings after you're finished collecting logs.

SavexG:
Thank you for trying to help

first to clear things up with fullscreen applications I mean applications that are not as convenient to close as firefox 
I meant mostly video games that run in fullscreen mode that you are not able to close or minimize with ease I hope you understand what I mean now

here is the first terminal output:

save@pop-os:~ $ xprop
[...]

PR #1125, opened 2021-08-03
Add Geometry Dash to skip taskbar exceptions

jacobgkau:
Intended to close https://github.com/pop-os/shell/issues/1123; needs tested by someone who owns Geometry Dash.

Not sure if this is the best long-term solution to the issue. If the issue affects many full-screen games, then better handling may be needed down the line.


mmstick (approved):

jacobgkau:
Closing this for now because user testing indicates it doesn't seem to fix the issue (and it seems all Proton games might be affected.)

Issue #1165, opened 2021-09-05
Can


senntore:
**(1) Issue/Bug Description:**
Cannot use TAB key to navigate

**(2) Steps to reproduce (if you know):**

**(3) Expected behavior:**

**(4) Distribution (run `cat /etc/os-release`):**
 

**(5) Gnome Shell version:**

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**

**(7) Where was Pop Shell installed from:**

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**

**(9) Other Installed/Enabled Extensions:**

**(10) Other Notes:**

jacobgkau:
Related: https://github.com/pop-os/shell/pull/797.

The tab key is used for autocompletion in the file browser plugin (`/` or `~`). The interaction with (or removal of) that feature would need to be considered when implementing this.

mmstick:
Maybe Ctrl + Tab could be used instead.

PR #1173, opened 2021-09-09
feat(launcher): Arrow key repeat on key press hold

mmstick:


jackpot51 (approved):

jacobgkau:
If I press and hold a key and the second-selected entry is a window, then that window stays highlighted even if I stop pressing while other entries are selected. Similarly, if I press and hold a key and the second-selected entry is not a window, then the window does not become highlighted if I actually stop on that window.

https://user-images.githubusercontent.com/7199422/132878054-ce122d66-633a-4537-91f7-2d0240e4468c.mp4

mmstick:
Fixed. Needed to add some select calls after the active ID is changed

jacobgkau:
The overlay is working as expected now; the window highlight overlay is always in sync with what's currently selected in the launcher.

Something else I noticed is that pressing Tab doesn't move to the next entry, but pressing and holding Tab does cycle through entries. Users have [requested](https://github.com/pop-os/shell/issues/1165) Tab to move to the next entry, but I thought that wasn't possible because some plugins use the Tab key. (Pressing and holding Tab in the file browser plugin completes one path, then starts cycling through entries instead of completing additional paths; that's also unexpected. If the key's repeating, a user would probably expect the behavior to be the same as it would be if the key was being pressed once.)

For consistency's sake, it seems like either pressing Tab once should move to the next entry, or pressing and holding it should not cycle through entries.

mmstick:
That's unusual. Something must have had handling for tab in here before.

jacobgkau (approved):
Tab no longer moves forward in the list when being held down. Shift-Tab still moves backwards (which is perhaps inconsistent, but matches current behavior for single presses.) Arrow keys and Ctrl-J/K still work as expected. Launcher regression testing passed on Hirsute and Focal.

PR #1179, opened 2021-09-14
fix: Disable config file monitoring due to GJS thread lock

### LINKED ISSUE #969: Pop-Shell freezes Gnome-Shell when using GJS 1.68.1

As soon as you start the [Manjaro 21.0.5 Review ISO](https://github.com/manjaro/release-review/releases/tag/202105140823) either in VirtualBox or baremetal Gnome-Shell will totally freeze when **Pop-Shell** is activated. We narrowed it down to **[GJS 1.68.1](https://gitlab.gnome.org/GNOME/gjs/-/tags/1.68.1)** to be the issue. Having the older version **1.68.0** running, it seems to be fine.

- Start the image as normal and boot into Gnome Shell
- It will freeze in under one minute, in Wayland session you can't even switch to another TTY

---

- Start the image but edit grub entry to boot into run level 3
- login as `manjaro:manjaro`
- install `downgrade` via: `sudo pacman -Sy downgrade`
- downgrade **gjs** to 1.68.0 via: `sudo downgrade gjs`
- either start gnome-shell with `startx` or `dbus-run-session -- gnome-shell --display-server --wayland`
- **Pop-Shell** won't freeze the shell

**(3) Expected behavior:**

**Pop-Shell** should not freeze with newer **gjs**

NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
BUILD_ID=rolling
PRETTY_NAME="Manjaro Linux"
ANSI_COLOR="32;1;24;144;200"
HOME_URL="https://manjaro.org/"
DOCUMENTATION_URL="https://wiki.manjaro.org/"
SUPPORT_URL="https://manjaro.org/"
BUG_REPORT_URL="https://bugs.manjaro.org/"
LOGO=manjarolinux

**(5) Gnome Shell version:**

1:3.38.4+13+gcf9d73ed5-1

Official Manjaro Repositories: 1.2.0-1

One Monitor in VirtualBox

Seems has nothing to do with other extensions .

**(10) Other Notes:**

We may need to debug this by rebuilding **gjs** and **js78** with [debug symbols](https://wiki.archlinux.org/title/GNOME/Troubleshooting#Shell_freezes) ...

mmstick:
Experienced GJS burning CPU the moment that I removed the local config files. Listening to `changed` events on a `Gio.File` seems to trigger the behavior.

Fixes #969

### DIFF
--- a/src/dialog_add_exception.ts
+++ b/src/dialog_add_exception.ts
@@ -15,7 +15,7 @@ export class AddExceptionDialog {
         shouldFadeOut: false
     });
 
-    constructor(cancel: () => void, this_app: () => void, current_window: () => void) {
+    constructor(cancel: () => void, this_app: () => void, current_window: () => void, on_close: () => void) {
         let title = St.Label.new("Add Floating Window Exception");
         title.set_x_align(Clutter.ActorAlign.CENTER)
         title.set_style("font-weight: bold")
@@ -34,6 +34,7 @@ export class AddExceptionDialog {
             label: "Cancel",
             action: () => {
                 cancel();
+                on_close()
                 this.close();
             },
             key: Clutter.KEY_Escape
@@ -43,6 +44,7 @@ export class AddExceptionDialog {
             label: "This App's Windows",
             action: () => {
                 this_app();
+                on_close()
                 this.close();
             },
         });
@@ -51,6 +53,7 @@ export class AddExceptionDialog {
             label: "Current Window Only",
             action: () => {
                 current_window();
+                on_close()
                 this.close();
             }
         })
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -554,20 +554,11 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     exception_select() {
-        if (GNOME_VERSION?.startsWith("3.36")) {
-            // 3.36 required a delay to work
-            GLib.timeout_add(GLib.PRIORITY_LOW, 500, () => {
-                this.exception_selecting = true
-                overview.show()
-                return false
-            })
-        } else {
-            GLib.idle_add(GLib.PRIORITY_LOW, () => {
-                this.exception_selecting = true
-                overview.show()
-                return false
-            })
-        }
+        GLib.timeout_add(GLib.PRIORITY_LOW, 500, () => {
+            this.exception_selecting = true
+            overview.show()
+            return false
+        })
     }
 
     exit_modes() {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -419,31 +419,6 @@ export class Ext extends Ecs.System<ExtEvent> {
         return window ? window.meta.get_compositor_private() : null;
     }
 
-    // TODO: Causes gjs to burn CPU cycles in a lock
-    // attach_config(): [any, SignalID] {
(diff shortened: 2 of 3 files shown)


jackpot51 (approved):

jackpot51:
@mmstick could you can make this branch master_impish so I can test new ISOs?

mmstick:
Pushed to `master_impish`

jacobgkau (changes requested):
When deleting a floating window exception, a previously auto-floated window no longer tiles. Even closing and re-opening the window does not make it tile. A shell restart is required to make the deleted rule no longer apply.

Likewise, after creating a new floating window exception, it no longer applies until the shell is restarted.

jackpot51:
This does appear to fix the hang on 21.10. I'm building new ISOs now. I'm not sure how to prevent the hang while also preventing the issue @jacobgkau described

jacobgkau:
> This does appear to fix the hang on 21.10. I'm building new ISOs now. I'm not sure how to prevent the hang while also preventing the issue @jacobgkau described

Maybe instead of listening for config changes, the config can at least be read when a user adds or removes a floating exception through the GUI?

Alternatively, to preserve more of the current behavior, would it be possible to check if the file exists first and only start watching if that's true? (Not sure if it works like that.)

mmstick:
The file is created before it is being watched. The first invocation of `this.conf.reload()` is in the constructor of the main `Ext` class. And the file watching signal is added in the `signals_attach` method that is called well after the `Ext` object was created.

Exceptions are added to the internal `Config`, but I can look at why they're not getting applied on a GUI change.

mmstick:
Let's try this: reload the config when the exceptions dialog is closed.

jacobgkau:
> Let's try this: reload the config when the exceptions dialog is closed.

That fixes creating a new exception, currently doesn't fix removing an exception or toggling built-in exceptions.

mmstick:
Okay. I've also added the reload for when the main exceptions application is closed.

jacobgkau:
> I've also added the reload for when the main exceptions application is closed.

That's not the same thing. Without immediately applying, it still appears to the user that deleting a manual exception or toggling a system exception didn't work. (Furthermore, a deleted exception still isn't reflected even if I do close out of the `Floating Window Exceptions` window on e3d9818.)

mmstick:
The exceptions dialog where you add and remove exceptions is a completely separate process so pop-shell is oblivious to what it's doing. It will take some time to come up with a mechanism for it to communicate to pop-shell to apply that change immediately.

mmstick:
Okay. The standard output of that process is now being watched asynchronously. `MODIFIED` signals will instruct pop-shell to reload the config.

jacobgkau:
On d2b9acb, I'm still seeing windows not re-tile when I delete their floating exceptions. This happens with Class or Class/Title exceptions. I'm testing with the GNOME Calculator app.

jacobgkau (approved):
d2b9acb seems to behave as expected. Regression testing passed:
[shell-pr1179-hirsute.txt](https://github.com/pop-os/shell/files/7179245/shell-pr1179-hirsute.txt)
[shell-pr1179-focal.txt](https://github.com/pop-os/shell/files/7179339/shell-pr1179-focal.txt)

Also confirmed in a live disk of 21.10 that this branch still fixes the hang issue when removing `~/.config/pop-shell/`.

PR #1196, opened 2021-10-01
Add tilda to default float-rules

jdtimmerman:
Tilda is a dropdown terminal, similar to guake which is already on this list.
The 'select window' interface of popshell didn't work for this app (it doesn't appear on the window overview) so adding it by default could save some people some time.


bflanagin (approved):
Works as expected, no regressions found.

PR #1171, opened 2021-09-09
fix(tiling): Windows launching onto window that isn't focused

mmstick:
Closes #1166

Issue #1183, opened 2021-09-17
Window opening on dual screen


LinuxHeki:
**(1) Issue/Bug Description:**
After an update (3 - 7 days ago) broke window opening on dual screen setup (just when you have tiling enabled). New window will always open on screen with already opened window. That happens also with Minecraft full screen mode. If I have opened Discord on one screen, Minecraft on other when I switch Minecraft to full screen mode it will open it on screen with Discord.
If I close Discord it will open it on the right screen.

**(2) Steps to reproduce (if you know):**
Open an app (browser), move it to any monitor, open another app (terminal) and it will open where another window is.

**(3) Expected behavior:**
Open new window / full screen mode in monitor where the mouse is / window is.

**(4) Distribution (run `cat /etc/os-release`):**
 NAME="Pop!_OS"
VERSION="21.04"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 21.04"
VERSION_ID="21.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=hirsute
UBUNTU_CODENAME=hirsute
LOGO=distributor-logo-pop-os

**(5) Gnome Shell version:**
Gnome 3.38.5 with Cosmic

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**

'''1.1.0~1631809642~21.04~5220303'''

**(7) Where was Pop Shell installed from:**
Apt, pre-installed

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**
2 x 1080p

**(9) Other Installed/Enabled Extensions:**
Blur my Shell,
GSConnect,
Sound Input & Output Device Chooser

**(10) Other Notes:**

jacobgkau:
> After an update (3 - 7 days ago) broke window opening on dual screen setup (just when you have tiling enabled).

The update may have been this PR: https://github.com/pop-os/shell/pull/1171 

> New window will always open on screen with already opened window.

This, in itself, is expected behavior.

> That happens also with Minecraft full screen mode. If I have opened Discord on one screen, Minecraft on other when I switch Minecraft to full screen mode it will open it on screen with Discord.

[...]

PR #1198, opened 2021-10-03
Undo fix(tiling) #1171

LinuxHeki:
Fix #1171 created issue #1183 so I'm fixing a fix that's not a fix. Sorry for a lot of fixes. I'm 14 and I don't know what am I doing. I just tried to fix an issue I found because  that issue is so annoying.


mmstick:
Reverting the prior fix will cause a severe regression in tiling behavior — windows will stop reliably attaching to the focused window. So it would be better to improve upon the fix instead.

PR #1205, opened 2021-10-06
fix(launcher): Exception in GNOME 3.36 due to `Shell.AppLaunchGPU` being undefined

### LINKED ISSUE #1194: Applications not launching when pressing enter in the launcher
When using the launcher applications will not start-up when selecting and pressing enter on them. Instead the window will partially freeze by not refreshing the results anymore (search and navigation still works), even though `pop-launcher` continuous to send new search results.

Diving deeper into the launcher code will reveal, that the issue is the unknown property `Shell.AppLaunchGpu` used in `launcher.ts#launch_desktop_entry` to assign `gpuPref`. Everything in the stack (shell, launcher, ...) up to this point works as expected. This will crash the function almost silently and as such the code to start the applications is not reached.  

`journalctl` will log this problem as the following, even though it is unclear to me when the error will be printed as it doesn't seem to be attached to the crashing of the above mentioned function
```
JS WARNING: [/home/elizabeth/.local/share/gnome-shell/extensions/pop-shell@system76.com/launcher.js 164]: reference to undefined property "AppLaunchGpu"
```

Probably using a non pop-os versioned gnome?

**(3) Expected behavior:**
Should launch

Default Ubuntu 20.04 with Wayland

**(5) Gnome Shell version:**
3.39.6

Pop-Shell: pop-os/shell@522030336565badc6fd6068bd5e38328996aa4bf
Pop-Launcher: pop-os/launcher@5cea11573dbb9180b144ac7420c0e6fba46aaefd

Manually

**(10) Other Notes:**
Replacing `Shell.AppLaunchGpu` with an empty string will fix this problem.

My recommendation would be to wrap the assignment of `gpuPref` into a try / catch block, or test beforehand whether the "needed" property exists in the gnome shell.

mmstick:
Closes #1194

### DIFF
--- a/src/launcher.ts
+++ b/src/launcher.ts
@@ -190,6 +190,11 @@ export class Launcher extends search.Search {
     }
 
     launch_desktop_entry(entry: JsonIPC.DesktopEntry) {
+        if (!("AppLaunchGpu" in Shell)) {
+            this.launch_gio(entry);
+            return;
+        }
+
         const basename = (name: string): string => {
             return name.substr(name.indexOf('/applications/') + 14).replace('/', '-')
         }
@@ -212,14 +217,7 @@ export class Launcher extends search.Search {
             log.error(`GNOME Shell cannot find desktop entry for ${desktop_entry_id}`)
             log.error(`pop-launcher will use Gio.DesktopAppInfo instead`);
 
-            const dapp = Gio.DesktopAppInfo.new_from_filename(entry.path);
-            
-            if (!dapp) {
-                log.error(`could not find desktop entry for ${entry.path}`);
-                return;
-            }
-
-            this.launch_desktop_app(dapp, entry.path);
+            this.launch_gio(entry)
             return;
         }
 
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

mmstick:
I'll look into that

matthewtraughber:
I'm **very** new to this, but I *think* these issues may be due to [gnome-shell](https://gitlab.gnome.org/GNOME/gnome-shell) switching from `gboolean` to `ShellAppLaunchGpu`.
- [3.36.8/shell-app.c](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/3.36.8/src/shell-app.c#L1342) `gboolean discrete_gpu` 
- [main/shell-app.c](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/src/shell-app.c#L1407) `ShellAppLaunchGpu gpu_pref`.

I manually updated and applied this fix locally to the latest `pop-os/shell` version, and got the launcher working again. I then modified things a bit (below) based on my **very** limited understanding - however, I don't have a discrete/default GPU config to test implementation.

```
        let gpuPref = null

        if (!("AppLaunchGpu" in Shell)) {
            gpuPref = false
        } else {
            gpuPref = entry.gpu_preference === "Default"
                ? Shell.AppLaunchGpu.DEFAULT
                : Shell.AppLaunchGpu.DISCRETE;
        }
```

PR #1206, opened 2021-10-06
feat: Improved HighContrast Support

mmstick:
highcontrast-ux

### DIFF
--- a/highcontrast.css
+++ b/highcontrast.css
@@ -1,0 +1,66 @@
+.pop-shell-active-hint {
+    border-style: solid;
+    border-color: #FBB86C;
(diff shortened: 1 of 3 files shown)


jacobgkau:
I'm still seeing a transparent/broken appearance while using HighContrast:

Screenshot from 2021-10-07 12-59-07

At first, I switched to HighContrast by setting the Applications theme in GNOME Tweaks, but I see the same thing if I enable it through Settings -> Accessibility. Is there some other way I should be activating HighContrast to see this?

mmstick:
Will switch to a Hirsute VM to test this

PR #1261, opened 2021-12-03
Fix Table of Contents

ikifar2012:
I noticed some broken links in the table of contents, so I have attempted to fix them


jacobgkau (approved):
Thank you for updating these links!

Issue #1128, opened 2021-08-05
Workspace related features broken


supermar1010:
**(1) Issue/Bug Description:**
Moving windows to a upper workspace does not work. Moving them a workspace down works without a problem. I checked the keybindings in the gnome keyboard settings dialog, there all the keybindings seem to be right. 

I guess a similiar source is my problem with moving focus from window to window. For example Super+Up  works, but Super+down does not work. Same goes for Super+right, which works, but Super+left, moves the current window to the left half of the screen (like it was in vanilla gnome). I also checked the keybindings here, but they seem to be right in my opinion. Any way to check these, with another tool than the gnome gui program.

**(2) Steps to reproduce (if you know):**

**(3) Expected behavior:**
Moving windows workspaces up should work.

**(4) Distribution (run `cat /etc/os-release`):**
NAME="Manjaro Linux"
ID=manjaro
ID_LIKE=arch
BUILD_ID=rolling
PRETTY_NAME="Manjaro Linux"
ANSI_COLOR="32;1;24;144;200"
HOME_URL="https://manjaro.org/"
DOCUMENTATION_URL="https://wiki.manjaro.org/"
SUPPORT_URL="https://manjaro.org/"
BUG_REPORT_URL="https://bugs.manjaro.org/"
LOGO=manjarolinux

**(5) Gnome Shell version:**
GNOME Shell 40.3

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**

commits: `ab87042d2579c9ad9bb3271584b0281d97da7baa`

**(7) Where was Pop Shell installed from:**

Arch aur, compiled from source `gnome-shell-extension-pop-shell-git`

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**

1x3440x1440

**(9) Other Installed/Enabled Extensions:**

**(10) Other Notes:**

supermar1010:
I just did some more experimenting, for the Switch Focus to window shortcuts, it worked reassigning them to the same shortcut used before, and then removing the custom assignment.
For the moving to different workspaces it did not work however.

For moving to different workspaces, I did not find the shortcut assigned in the gnome settings so I set it myself, this works now. So it seems for me the setting of the keyboard shortcuts went wrong?

RichardSieg:
[...]

PR #1262, opened 2021-12-04
Fix workspace moving in Gnome 40

dpbackes:
- In Gnome 40, workspaces are organized horizontally rather than
  vertically. Thus, we need to handle moving left and right as well as 
  up and down

Fixes https://github.com/pop-os/shell/issues/1128

-------

This seems to be a move in the right direction but there are still some bugs. It works most of the time but, occasionally, moving a window will stop working. Usually in only one direction (ie, I'll be able to move windows to the right workspace, but not back to the left or vice versa). When this happens, it's because `move_workspace` is not being called. 

I'll continue debugging on my end but I have zero experience in gnome extension development so I'm putting this up with the thought that others who are interested in #1128 might have more success than me in figuring out the remaining problems.

### DIFF
--- a/keybindings/10-pop-shell-move.xml
+++ b/keybindings/10-pop-shell-move.xml
@@ -8,6 +8,8 @@
   <KeyListEntry name="tile-move-up" description="Move window up"/>
   <KeyListEntry name="pop-workspace-down" description="Move window to lower workspace"/>
   <KeyListEntry name="pop-workspace-up" description="Move window to upper workspace"/>
+  <KeyListEntry name="pop-workspace-left" description="Move window to left workspace"/>
+  <KeyListEntry name="pop-workspace-righ" description="Move window to right workspace"/>
   <KeyListEntry name="pop-monitor-down" description="Move window to lower monitor"/>
   <KeyListEntry name="pop-monitor-left" description="Move window to leftward monitor"/>
   <KeyListEntry name="pop-monitor-right" description="Move window to rightward monitor"/>
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -202,6 +202,16 @@
             <summary>Move window to the upper workspace</summary>
         </key>
 
+        <key type="as" name="pop-workspace-righ">
+            <default><![CDATA[['<Super><Shift>Right','<Super><Shift>KP_Right','<Super><Shift>l']]]></default>
+            <summary>Move window to the right workspace</summary>
+        </key>
+
+        <key type="as" name="pop-workspace-left">
+            <default><![CDATA[['<Super><Shift>Left','<Super><Shift>KP_Left','<Super><Shift>h']]]></default>
+            <summary>Move window to the left workspace</summary>
+        </key>
+
         <key type="as" name="pop-monitor-down">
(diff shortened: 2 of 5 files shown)


keybindings/10-pop-shell-move.xml
```diff
@@ -8,6 +8,8 @@
   <KeyListEntry name="tile-move-up" description="Move window up"/>
   <KeyListEntry name="pop-workspace-down" description="Move window to lower workspace"/>
   <KeyListEntry name="pop-workspace-up" description="Move window to upper workspace"/>
+  <KeyListEntry name="pop-workspace-left" description="Move window to left workspace"/>
+  <KeyListEntry name="pop-workspace-righ" description="Move window to right workspace"/>
```
jacobgkau, line 12: Based on the names of other keys, this should probably be `pop-workspace-right` instead of `pop-workspace-righ`.

schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
```diff
@@ -202,6 +202,16 @@
             <summary>Move window to the upper workspace</summary>
         </key>
 
+        <key type="as" name="pop-workspace-righ">
```
jacobgkau, line 205: These seem to take effect in a horizontal layout as well, which probably isn't desired behavior.

scripts/configure.sh
```diff
@@ -74,6 +74,10 @@ set_keybindings() {
     dconf write ${KEYS_GNOME_WM}/switch-to-workspace-down "['<Primary><Super>Down','<Primary><Super>${down}']"
     # Move to workspace above
     dconf write ${KEYS_GNOME_WM}/switch-to-workspace-up "['<Primary><Super>Up','<Primary><Super>${up}']"
+    # Move to workspace left
```
jacobgkau, line 77: These keybindings would conflict with half-snapping; I see half-snapping was already disabled in this file, although it's still present in `TESTING.md` and is something we currently test for.

jacobgkau (commented):
This may need more design consideration (last time I remember this being discussed as https://github.com/pop-os/shell/pull/805.)

Frequential:
Is this still being discussed? I'm assuming it's not of a high priority since Pop!_OS still isn't using GNOME 40.

jacobgkau:
> Is this still being discussed? I'm assuming it's not of a high priority since Pop!_OS still isn't using GNOME 40.

Pop!_OS actually is using GNOME 40 as of today (21.10), but it uses the vertical workspaces extension.

Frequential:
> > Is this still being discussed? I'm assuming it's not of a high priority since Pop!_OS still isn't using GNOME 40.
> 
> Pop!_OS actually is using GNOME 40 as of today (21.10), but it uses a vertical workspaces extension.

Oh, I wasn't aware of that, there's a possibility this never gets merged then I guess.

dpbackes:
Ahh well that makes sense as a solution! Thanks for everyone's attention, really appreciate the project.

Frequential:
Wait! This is still causing problems in every other non Pop!_OS distro using GNOME or simply people activating horizontal workspaces.

michalvankodev:
It would be great to link the extension in the installation for different OSes. It looks like it's an essential part of the experience. I just discovered this and I <3 this very much ! Something I was trying to find for many years.

Issue #1264, opened 2021-12-05
pop-shell@system76.com/extensions.js" prevents Alt-TAB shortcut (switch application) from working


abrighton:
**(1) Issue/Bug Description:**

After installing updates, ALt-TAB keyboard shortcut (switch application) stopped working and syslog or journalctl
showed an error at line 2124 of /usr/share/gnome-shell/extensions/pop-shell@system76.com/extension.js: "app is NULL".
The line in the original source (extension.ts) is 2728.

The original line was:

                        return app.get_name() === w_app.get_name();

The fix is to add a check:

                        return app ? app.get_name() === w_app.get_name() : false;

**(2) Steps to reproduce (if you know):**

Type Alt-TAB to switch applications.

**(3) Expected behavior:**

Switch applications.

**(4) Distribution (run `cat /etc/os-release`):**
 
NAME="Pop!_OS"
VERSION="21.04"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 21.04"
VERSION_ID="21.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=hirsute
UBUNTU_CODENAME=hirsute
LOGO=distributor-logo-pop-os

**(5) Gnome Shell version:**

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**

```
pop-shell:
  Installed: 1.1.0~1638556835~21.04~b2cfd69
  Candidate: 1.1.0~1638556835~21.04~b2cfd69
  Version table:
 *** 1.1.0~1638556835~21.04~b2cfd69 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu hirsute/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu hirsute/main i386 Packages
        100 /var/lib/dpkg/status
```

**(7) Where was Pop Shell installed from:**

system

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**

**(9) Other Installed/Enabled Extensions:**

**(10) Other Notes:**

mmstick:
Please create a PR with the change.

jacobgkau:
[...]

PR #1265, opened 2021-12-05
GitHub issue #1264: Fix for Null "app"

abrighton:
GitHub issue #1264: Fix for Null "app", which prevents Alt-TAB shortcut (switch application) from working


jmmaranan:
You can also add null check for `w_app`. It could be null as well

jacobgkau (approved):
Confirmed this fixes https://github.com/pop-os/shell/issues/1264 and gets the Alt-Tab window switcher working when the LibreOffice Document Recovery dialog is open.

Regression testing passed:
[shell-pr1265-impish.txt](https://github.com/pop-os/shell/files/7670087/shell-pr1265-impish.txt)
[shell-pr1265-focal.txt](https://github.com/pop-os/shell/files/7670421/shell-pr1265-focal.txt)

Issue #1278, opened 2021-12-17
Lost ability to close the windows via ctrl+q in the launcher


dodalovic:
If I recall correctly, once I press the Super key, and if I want to close the highlighted app, I was able to do it previously via `ctrl+q` combo. 
It doesn't work anymore. 

Any thoughts?
Thanks for the wonderful product!

mmstick:
That was never supported. The Super key is always held with OS shortcuts. Ctrl + Q would be an application-level shortcut that not every application is going to support.

dodalovic:
https://github.com/pop-os/shell/issues/857#issuecomment-829093307

13r0ck:
Closed via pop-os/launcher#75

PR #1286, opened 2021-12-21
fix: Close search window on 'WindowQuit'

ningvin:
This is related to #1278 and [my pull request on pop-os/launcher](https://github.com/pop-os/launcher/pull/75).

When receiving a call to `WindowQuit`, the launcher window currently stays open after the requested window is closed. The entry list in the launcher window is not updated however, with the entry for the now closed window still being present.

This change causes the launcher window to close after each `WindowQuit` call, which should be the same behavior as was implemented [before adapting the Rust launcher implementation](https://github.com/pop-os/shell/blob/e83443e96ba914f3ef990e3e410a0dbe39f11c58/src/dialog_launcher.ts#L312).

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -262,6 +262,7 @@ export class Ext extends Ecs.System<ExtEvent> {
 
         this.dbus.WindowQuit = (win: [number, number]) => {
             this.windows.get(win)?.meta.delete(global.get_current_time())
+            this.window_search.close()
         }
     }
 


mmstick (approved):

jacobgkau (approved):
This along with https://github.com/pop-os/launcher/pull/75 restores window closing functionality. Regression testing passed:
[shell-pr1286-impish.txt](https://github.com/pop-os/shell/files/7765044/shell-pr1286-impish.txt)
[shell-pr1286-focal.txt](https://github.com/pop-os/shell/files/7765202/shell-pr1286-focal.txt)

PR #1336, opened 2022-01-24
add re.sonny.Junction to DEFAULT_FLOAT_RULES

bryanpaget:
The [Junction app](https://flathub.org/apps/details/re.sonny.Junction) for choosing a web browser when clicking on links should be excluded from tiling by default.

Also: should this list be alphabetical?


jacobgkau (approved):
Confirmed Junction does not tile, and that this exception works on Pop!_OS 21.10 and 20.04 LTS. The exception being added is the only change being made.

> Also: should this list be alphabetical?

I don't think the order makes any functional difference, but it probably wouldn't be a bad idea.

PR #1339, opened 2022-01-28
feat(exceptions): Add "Mozilla VPN"

bryanpaget:
1. Add `"Mozilla VPN"` to `DEFAULT_FLOAT_RULES`.
2. Sort array entries alphabetically.
3. Remove trailing commas, for instance: `{ class: "gnome-screenshot", } --> { class: "gnome-screenshot" }`


jacobgkau (approved):
Confirmed Mozilla VPN did not previously tile properly, and now floats by default. Also confirmed that existing exceptions, both class and class/title, are still working. Tested on Pop!_OS 21.10 and 20.04 LTS.

PR #1357, opened 2022-03-02
fix: Do not unstack stack on last remaining window

### LINKED ISSUE #1301: Last stacked window gets unstacked automatically
When I close out of the second to last window in a stack the last remaining window is left unstacked. This is very aggravating as I have to create a new stack every time I open a new window. Otherwise the new window is not stacjed and is instead tiled as it would be normally. 

Here is a previous issue citing the same occurrence:

I can confirm this; the last window remains in stack mode if you remove the penultimate window from the stack, however closing that window destroys the stack. 

Unfortunately the commit doesn't fix the issue. 

-- bvs

_Originally posted by @bettyvschmartz in https://github.com/pop-os/shell/issues/881#issuecomment-886538473_

mmstick:
Closes #1301

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -232,9 +232,9 @@ export class AutoTiler {
     }
 
     /** Detaches the window from a tiling branch, if it is attached to one. */
-    detach_window(ext: Ext, win: Entity, destroy_stack: boolean = true) {
+    detach_window(ext: Ext, win: Entity) {
         this.attached.take_with(win, (prev_fork: Entity) => {
-            const reflow_fork = this.forest.detach(ext, prev_fork, win, destroy_stack);
+            const reflow_fork = this.forest.detach(ext, prev_fork, win);
 
             if (reflow_fork) {
                 const fork = reflow_fork[1];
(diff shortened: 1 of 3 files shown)


jackpot51 (dismissed):

jacobgkau:
The behavior's now inconsistent when dragging a window out of a stack with the mouse. When the window is dropped into the same tree, then the remaining window is still unstacked; when the window is dropped into a different tree, then the remaining window is not unstacked.

https://user-images.githubusercontent.com/7199422/156460032-2e051642-56b0-4b90-939b-e99c387f12ac.mp4

Is there any way to make this more predictible? If we're wanting to not unstack the last window, then it shouldn't be getting unstacked in any of the three operations shown above.

mmstick:
@jacobgkau Fixed it. A drag out of a stack won't auto-unstack the remaining window now.

jackpot51 (approved):

jacobgkau (changes requested):
df0ba04 fixed the issues I showed in the video before; the stack is no longer destroyed in any of those cases.

However, I am not able to destroy a stack with Super-S, which is a bigger regression. Once I put a window into a stack using mouse or keyboard, the stack is stuck there. Trying to move the last window out of the stack with window management mode + arrow keys also does not work.

mmstick:
Okay. Found the correct code to remove to keep unstacks by keyboard and management mode working.

PR #1358, opened 2022-03-02
fix: On destroy of stacked window, activate previously active tab

mmstick:
Useful when a web browser has a popup window and then that popup is closed

Or when testing GTK applications from an IDE

### DIFF
--- a/src/forest.ts
+++ b/src/forest.ts
@@ -717,18 +717,15 @@ export class Forest extends Ecs.World {
 
             // Activate the next window in the stack if the window was destroyed.
             if (idx !== null) {
-                let activate = idx > 0 ? idx - 1 : 0;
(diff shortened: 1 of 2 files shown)


jackpot51 (dismissed):

jacobgkau (approved):
Previously, the last tab in a stack would be activated if the active window was closed. Now, the most recently active tab is activated. (This also gets rid of a flicker that used to occur before the last tab was activated where the most recent tab's window was shown instead of the last tab's window.)

Regression testing passed:
[shell-pr1358-impish.txt](https://github.com/pop-os/shell/files/8187965/shell-pr1358-impish.txt)
[shell-pr1358-focal.txt](https://github.com/pop-os/shell/files/8188019/shell-pr1358-focal.txt)

PR #1360, opened 2022-03-02
fix(launcher): Hide previous popup menu if another is shown

### LINKED ISSUE #1359: "Launch using discrete graphics" prompt can be spawned multiple times
Screenshot from 2022-03-02 12-41-46

I saw this when I found https://github.com/pop-os/launcher/issues/89, but it's also appearing on a switchable machine while in Hybrid mode, which is when the prompt should be appearing (though presumably only once).

mmstick:
Closes #1359

### DIFF
--- a/src/launcher.ts
+++ b/src/launcher.ts
@@ -28,6 +28,7 @@ export class Launcher extends search.Search {
     windows: arena.Arena<ShellWindow> = new arena.Arena()
     service: null | service.LauncherService = null
     append_id: null | number = null
+    active_menu: null | any = null
 
     constructor(ext: Ext) {
         super()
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

jacobgkau (approved):
Confirmed this prevents multiple Discrete Graphics popups from showing at the same time (when a new one is opened, the old one closes.)

Launcher regression testing passed:
[shell-pr1360-impish.txt](https://github.com/pop-os/shell/files/8187028/shell-pr1360-impish.txt)
[shell-pr1360-focal.txt](https://github.com/pop-os/shell/files/8187049/shell-pr1360-focal.txt)

`launcher.ts` is the only file being modified.

PR #1361, opened 2022-03-03

Blue-EyesChaosMaxDragon:
If extension is installed in ./local/share/gnome-shell/extensions/ make sure to enable user extensions

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -14,7 +14,7 @@ Depends:
     pop-launcher,
     pop-shell-shortcuts,
     fd-find
-Recommends: pop-shell-plugin-system76-power
+Recommends: pop-shell-plugin-system76-power, system76-scheduler
 Replaces: gnome-control-center-data (<< 1:3.38.1-2ubuntu1pop1~)
 Breaks: gnome-control-center-data (<< 1:3.38.1-2ubuntu1pop1~)
 Description: Pop!_OS Shell
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -24,6 +24,7 @@ import * as stack from 'stack';
 import * as add_exception from 'dialog_add_exception';
 import * as exec from 'executor';
 import * as dbus_service from 'dbus_service';
+import * as scheduler from 'scheduler';
 
 import type { Entity } from 'ecs';
 import type { ExtEvent } from 'events';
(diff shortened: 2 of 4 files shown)


jacobgkau (commented):
Thank you for the PR. I can confirm this works to enable user extensions when installing with `make local-install` on a system where they were previously disabled.

Is there a reason you modified the makefile instead of putting this in `configure.sh`? I'm just wondering because it looks like the makefile was trying to only call that script, and the script already includes other dconf changes.

Blue-EyesChaosMaxDragon:
What is the purpose of an installation with the `DESTDIR` variable set?
I think `/org/gnome/shell/disable-user-extensions` only effects extensions in `~/.local/share/gnome-shell/extensions`. So I wanted to apply the dconf change only if `~/.local/share/gnome-shell/extensions` is the `INSTALLBASE`. To achieve this, I put it in the makefile and used the same if condition as in line 9 of the makefile. I am not an expert, so please correct me if I am wrong.

mmstick:
`DESTDIR` is used by debian packaging

Blue-EyesChaosMaxDragon:
So no need to change `/org/gnome/shell/disable-user-extensions`  if `DESTDIR` is used right?
One could move the if condition and the dconf change into the `configure.sh`, but would need some additional lines to make it work.

mmstick:
DESTDIR is only used by the temporary chroot that builds the debian packages.

Blue-EyesChaosMaxDragon:
Okay, if debian packing is used, the pop-shell extension is installed to  `/usr/share/gnome-shell/extensions` and not to `~/.local/share/gnome-shell/extensions`, right? So there would be no need to apply the dconf change.

mmstick:
No, it goes to `${PWD}/debian/tmp/usr/share/`

Blue-EyesChaosMaxDragon:
So is the if condition necessary or not? Sorry for all the questions, I am nor very familiar with debian packaging.

mmstick:
It is not. The default value is `''` unless it's being packaged by dpkg-buildpackage/sbuild

Blue-EyesChaosMaxDragon:
Okay so if I install via a package, the package still ends in `~/.local/share/gnome-shell/extensions`?
If yes, I close this pull request and create a new one with just the dconf change in the `configure.sh`

mmstick:
Packages install to `/` and they don't run commands to change shortcuts on extraction.

Blue-EyesChaosMaxDragon:
Thank you for all your explanations, I will open a new pull request without the if condition

PR #1363, opened 2022-03-06
chore: Mark Discord initial updater pop-up as floating

Xanewok:
Using just `title` seems more idiomatic, but I can see that other exceptions keep using `class` to refer to titles? Let me know if I should change that.


jacobgkau (approved):
Float exception is working. Before:
Screenshot from 2022-03-07 10-05-00

After:
Screenshot from 2022-03-07 10-12-20

Discord still tiles once the updater goes away.

jackpot51 (approved):

PR #1364, opened 2022-03-07
fix: entity is null on window destroy from stack

mmstick:
Fixes regression from last commit that causes tiling to stop working when an entity null error is triggered. It occasionally happens when two stacks are active, windows are added to both, and focus is shifted back and forth while removing and adding windows.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -747,6 +747,8 @@ export class Ext extends Ecs.System<ExtEvent> {
         const window = this.windows.get(win);
         if (!window) return;
 
+        window.destroying = true;
+
         // Disconnect all signals on this window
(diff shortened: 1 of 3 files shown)


jackpot51 (approved):

jacobgkau (approved):
Haven't recreated the issue (also didn't try for too long, but I did create two stacks, then switch focus between them while opening/closing windows), but regression testing has passed:
[shell-pr1364-impish.txt](https://github.com/pop-os/shell/files/8199730/shell-pr1364-impish.txt)
[shell-pr1364-focal.txt](https://github.com/pop-os/shell/files/8199873/shell-pr1364-focal.txt)

PR #1366, opened 2022-03-07
fix: On destroy of stacked window, activate previously active tab

mmstick:
Can be tested by launching an application from a terminal, and then closing the application should switch back to the terminal that spawned it, regardless of its position in the stack.

### DIFF
--- a/src/forest.ts
+++ b/src/forest.ts
@@ -713,31 +713,16 @@ export class Forest extends Ecs.World {
             this.stacks.remove(stack.idx)?.destroy();
             on_last();
         } else {
-            const idx = Node.stack_remove(this, stack, window);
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

jacobgkau (approved):
This seems to restore the behavior described in https://github.com/pop-os/shell/pull/1358#pullrequestreview-900565221.

Regression testing passed:
[shell-pr1366-impish.txt](https://github.com/pop-os/shell/files/8208553/shell-pr1366-impish.txt)
[shell-pr1366-focal.txt](https://github.com/pop-os/shell/files/8208665/shell-pr1366-focal.txt)

PR #1367, opened 2022-03-08
chore: Mark Ubuntu Software Updater as floating window

Xanewok:
Similar to #1363

The name seems to be generic enough that it probably warrants specifying the class as well, but I'm no GTK developer and I have no idea how to get app id from a window on window select/hover... could you help me by pointing me in the right direction?

### Reproduction
Run `update-manager`

### Screenshots

Before:
Screenshot from 2022-03-08 13-18-08

After:
Screenshot from 2022-03-08 13-17-33

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -36,6 +36,7 @@ export const DEFAULT_FLOAT_RULES: Array<FloatRule> = [
     { class: "Io.elementary.sideload" },
     { class: "KotatogramDesktop", title: "Media viewer" },
     { class: "Mozilla VPN" },
+    { title: "Software Updater" },
     { class: "Solaar" },
     { class: "Steam", title: "^((?!Steam).)*$" },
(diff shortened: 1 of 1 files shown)


mmstick:
`xprop` then click the window; or use the exceptions select button from the floating window exceptions dialog to choose the window and add a rule for the whole application.

Xanewok:
Awesome, thank you!

> WM_CLASS(STRING) = "update-manager", "Update-manager"

That's what I'm looking for, right? Should I add this class there, instead?

mmstick:
Yeah. They would be better

Xanewok:
Done!

jacobgkau (approved):
Confirmed on Ubuntu that the Software Updater app does not tile properly (window cannot be resized), and that the floating exception on this branch works.

PR #1368, opened 2022-03-09

Blue-EyesChaosMaxDragon:
If the extension is installed with `make local-install` to `~/./local/share/gnome-shell/extensions/`, it is now ensured that user extensions are enabled.
Improved version of pull request [#1361](https://github.com/pop-os/shell/pull/1361)

### DIFF
--- a/scripts/configure.sh
+++ b/scripts/configure.sh
@@ -106,6 +106,9 @@ fi
 
 set_keybindings
 
+# Make sure user extensions are enabled
+dconf write /org/gnome/shell/disable-user-extensions false
+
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Testing on Ubuntu 22.04 with both Wayland and X11, I still get `Extension "pop-shell@system76.com" does not exist` at the end of `make local-install`, and I still have to restart GNOME Shell to get it to show up in the Extensions app for me to enable. However, this change does work to enable user extension (the `Extensions` switch in the Extensions app), which is necessary before I can manually enable Pop Shell.

PR #1370, opened 2022-03-11
Re-apply window title on/off after screen unlocking

fhackenberger:
Before this change the window titles were still visible after
unlocking the screen, even if the setting was to off.

### DIFF
--- a/src/dbus_service.ts
+++ b/src/dbus_service.ts
@@ -25,6 +25,7 @@ const IFACE: string = `<node>
 export class Service {
     dbus: any
     id: any
+    handlerScsvWakeUp: any; // Handler id from DBus signal subscription (need to unsubscribe on destroy)
 
     FocusLeft: () => void = () => {}
     FocusRight: () => void = () => {}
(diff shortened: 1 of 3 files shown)


mmstick:
This adds methods to pop-shell's public DBus API. Did you mean to add these methods to the extension class instead?

fhackenberger:
> I'm not able to recreate this issue on Pop!_OS 21.10, 20.04 LTS, or 22.04 LTS pre-release. 
I'm using Ubuntu 20.04 LTS with pop-shell compiled from git master and experience it with Thunderbird for example. Most gnome apps are unaffected these days as they haven't got separate window titles anymore. I lock my screen with Ctrl-Alt-L or suspend the laptop.

> This adds methods to pop-shell's public DBus API. Did you mean to add these methods to the extension class instead?
I was unsure where a DBus signal subscription would fit best. I didn't add this method to the DBus interface, so it's not a publicly (over DBus) callable method and prefixed with on... to indicate that it's an event handler. Do you think it should rather be in the Ext class?

PR #1372, opened 2022-03-15
🔘 Increased corner radius size

tanaybhomia:
People use a lot of different themes on different distros and as the gnome team is increasing the corner size themes have also been increasing their sizes and due to this the window border looks weird when under a theme like per say default and custom one. 

I also was not able to use 2 to 3 themes because of this. 

I hope you find this use

### DIFF
--- a/dark.css
+++ b/dark.css
@@ -1,7 +1,7 @@
 .pop-shell-active-hint {
     border-style: solid;
     border-color: #FBB86C;
-    border-radius: 5px;
+    border-radius: 6px;
     box-shadow: inset 0 0 0 1px rgba(24, 23, 23, 0)
 }
 
--- a/dark.css
+++ b/dark.css
@@ -10,11 +10,11 @@
 }
 
 .pop-shell-border-normal {
-    border-width: 3px;
+    border-width: 2px;
 }
 
 .pop-shell-border-maximize {
-    border-width: 3px;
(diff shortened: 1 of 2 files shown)


jacobgkau (commented):
Before:
Screenshot from 2022-03-15 10-40-18

After:
Screenshot from 2022-03-15 10-41-04

This isn't just increasing the border radius, it's also decreasing the thickness. I could see a case for thickness being configurable (similar to how color is currently configurable), but the default seems significantly less easy to see at a glance after the change.

@Deottive Can you please revert the border size change and leave only the radius change? The new radius seems to match windows in the Pop theme better, so it may be an acceptable change by itself, although it will still need UX review (CC @maria-komarova).

jacobgkau (changes requested):
After speaking with @maria-komarova about the active hint radius change, we found that the current value of 5px is one pixel greater than the [4px window radius set by the Pop theme](https://github.com/pop-os/gtk-theme/blob/master/gtk/src/light/gtk-3.0/_common.scss), and we don't want the active hint radius changed to 6px because that is getting farther away from our window radius.

However, if you can make the active hint size set based on the GTK theme instead of being hard-coded, we would be open to that. (This would be `$window_radius` + 1px if we're matching the current active hint radius.) That would allow you to have a greater active hint radius with themes that use a greater window radius.

PR #1374, opened 2022-03-22
chore: add Pritunl client to floating exceptions

Yethal:
Pritunl desktop is a non-resizeable electron app. Adding it via Floating Window Exceptions works fine but this will save people a config step


mmstick (approved):

jacobgkau (approved):
Confirmed the Pritunl OpenVPN client does not tile properly (its size is somewhat resizeable, but with a maximum width and height that makes tiling ineffective.)

Also confirmed this floating exception works (built locally to test).

PR #1377, opened 2022-03-25
Add gnome 42 to supported list

MichaelAquilina:
Been testing this for about an hour now and seems to work exactly as expected :)


mmstick:
You must use the `master_jammy` branch for required GNOME 42 fixes.

PR #1378, opened 2022-03-25
fix: Gap buttons not functional in GNOME 42

mmstick:


jackpot51 (approved):

jacobgkau (approved):
The hitbox is a little unclear (which I assume has to do with the shell theme not working properly), but this gets the `+`/`-` buttons for gap adjustment working on 22.04.

Regression testing passed:
[shell-pr1378-jammy.txt](https://github.com/pop-os/shell/files/8353884/shell-pr1378-jammy.txt)

Issue #1386, opened 2022-03-31
Gnome 42: Click outside Launcher doesn't close it


ids1024:
Same issue as https://github.com/pop-os/cosmic/issues/324.

As I understand it, https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/2045 and have changing things so that input events no longer propagate to parents of a dialog. This seems to break the behavior of `shellReactive` dialogs as used in a few places here.

PR #1389, opened 2022-04-01
[jammy] fix: On Gnome 42, close launcher on click outside of it

ids1024:
Fixes https://github.com/pop-os/shell/issues/1386.

This breaks the behavior on older Gnome Shell versions. If we want this branch to work there as well, it would be easy to make it conditional.

I don't know if there are any other undesired behavior changes here or the other places `shellReactive: true` is used, but it seems to basically work as expected now.

### DIFF
--- a/src/search.ts
+++ b/src/search.ts
@@ -206,7 +206,7 @@ export class Search {
         // Ensure that the width is at least 640 pixels wide.
         this.dialog.contentLayout.width = Math.max(Lib.current_monitor().width / 4, 640);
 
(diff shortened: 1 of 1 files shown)


mmstick (approved):

n3m0-22 (approved):
Tested and working correctly on GNOME 40.5. I see no other undesired behaviors.

leviport (changes requested):
Need to test on 22.04 (which should have Gnome 42)

n3m0-22 (approved):
Working properly in 22.04 GNOME 42

PR #1408, opened 2022-04-29
fix: Windows not activated on workspace switch

### LINKED ISSUE #1407: No Window has Focused if Moving From an Empty Workspace
To reproduce:

1. Launch an app
2. Use shortcuts to move to the next empty workspace
3. Use shortcuts to move back up
4. The window does not have focus

The problem with this behavior is that Super + HJKL or Enter can no longer change focus or activate a window. The user has to click the window with the cursor.

mmstick:
Closes #1407

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -170,6 +170,9 @@ export class Ext extends Ecs.System<ExtEvent> {
 
     private size_requests: Map<GObject.Object, SignalID> = new Map();
 
+    /** Stores windows that were focused on a workspace */
+    private workspace_active: Map<number, null | Entity> = new Map()
+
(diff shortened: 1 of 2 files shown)


n3m0-22 (approved):
I was able to reproduce the issue. After installing this I tested it with and without Show Active Hint and with and without Tile Windows. The windows now stay active when switching between workspaces and can be selected and focus changed with keyboard shortcuts.

linuxgnuru:
Could this be upstream from gnome?

 https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/5410

13r0ck (approved):

PR #1416, opened 2022-05-05

jacobgkau:
Applies https://github.com/pop-os/shell/pull/1368 to Jammy, since it was merged into master for older releases.

Note for QA, this `configure.sh` script is only used for local installations, not for the official Pop!_OS .deb packaging.

### DIFF
--- a/scripts/configure.sh
+++ b/scripts/configure.sh
@@ -106,6 +106,9 @@ fi
 
 set_keybindings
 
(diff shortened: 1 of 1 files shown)


13r0ck (approved):

n3m0-22 (approved):
This correctly enables the user-extension switch.

Issue #1415, opened 2022-05-05
Focus is sometimes lost when moving window between workspaces


jmbuhr:
**Distribution (run `cat /etc/os-release`):**
```
NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=jammy
UBUNTU_CODENAME=jammy
LOGO=distributor-logo-pop-os
```

**Issue/Bug Description:**
I previously encountere this issue only when activating "focus on hover", but now it appears to be back and this time it also appears with the default "click to focus" (https://github.com/pop-os/shell/issues/550)

**Steps to reproduce (if yoYu know):**
With multiple windows open on multiple workspaces attempt to move one window from the last to the first workspace by repeatedly pressing `shit+super+k/j`

**Expected behavior:**
Window stays focussed so that the same window is moved at every step.

**Other Notes:**

Mindus-Amitiel-Debsin:
I also have this issue - trying to move a window down multiple workspaces seems to be difficult and often the window will lose focus to another window.
I have my workspace movement buttons as L-Ctrl + Pg_Up to move one workspace up, and L-Ctrl + Pg_Dwn to move one workspace down. For moving windows, I use the Shift modifier on the workspace movement commands. 
I have found that using the shortcuts will also redirect my cursor to the top-left of the window, for example my Chrome (top-left option) close and minimize buttons, or on the top-left corner of my Steam friend's list.

While replicating this issue on screencast, I found that of three times I tried, the window only lost focus when I was going down Workspaces, IE with L-Shift + L-Ctrl + Pg_Dwn. 

I am using Pop!_OS 22.04 Intel version and just did a fresh install of it this morning.
Here is the 6MB Screencast video: https://drive.google.com/file/d/1KV65QeV33-eBdEzhW85s07TBazaOe94H/view?usp=sharing

> NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
[...]

PR #1417, opened 2022-05-05
Feature: Add option to disable moving mouse cursor on keyboard navigation

### LINKED ISSUE #1188: move_pointer_on_switch=false doesn't work
- `"move_pointer_on_switch": false` in `config.json` has no effect

- Open `config.json`, set `move_pointer_on_switch` to `false
- Open two windows
- Switch between the two windows and note that the pointer _does_ move

**(3) Expected behavior:**
- The pointer stays where it was last moved _manually_ with the mouse

```
$ cat /etc/os-release
NAME=Fedora
VERSION="34 (Thirty Four)"
ID=fedora
VERSION_ID=34
VERSION_CODENAME=""
PLATFORM_ID="platform:f34"
PRETTY_NAME="Fedora 34 (Thirty Four)"
```

**(5) Gnome Shell version:**
- `40.4.0` running `X`

- Not 100% sure how to obtain this when built from source (code is cloned into a `mktemp` directory and removed later).

- Built from source

- Single, `3840x1600`

image

**(10) Other Notes:**

AKK9:
This is my attempt at implementing #1188.

Do I need to add a new scenario to TESTING.md?

Does the setting summary `Move cursor to active window when switching via keyboard shortcuts` accurately describe the behaviour here?

Also looking for suggestions on the label for this setting in the settings window, it doesn't make much sense at the moment.

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -27,6 +27,11 @@
             <summary>Handle minimized to tray windows</summary>
         </key>
 
+        <key type="b" name="move-cursor-on-switch">
+            <default>true</default>
+            <summary>Move cursor to active window when switching via keyboard shortcuts</summary>
+        </key>
+
         <!-- Tiling Options -->
         <key type="u" name="column-size">
             <default>64</default>
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -27,9 +27,9 @@
             <summary>Handle minimized to tray windows</summary>
         </key>
 
-        <key type="b" name="move-cursor-on-switch">
+        <key type="b" name="mouse-cursor-follows-active-window">
             <default>true</default>
-            <summary>Move cursor to active window when switching via keyboard shortcuts</summary>
+            <summary>Move cursor to active window when navigating with keyboard shortcuts or touchpad gestures</summary>
         </key>
 
         <!-- Tiling Options -->
--- a/src/config.ts
+++ b/src/config.ts
@@ -101,9 +101,6 @@ export class Config {
     /** Logs window details on focus of window */
     log_on_focus: boolean = false;
 
-    /** Move pointer when you switch applications */
-    move_pointer_on_switch: boolean = false;
-
     /** Specify default pointer position when you're switching windows */
     default_pointer_position: DefaultPointerPosition = DefaultPointerPosition.TopLeft;
 
@@ -180,7 +177,6 @@ export class Config {
             this.float = c.float;
             this.log_on_focus = c.log_on_focus;
             this.default_pointer_position = c.default_pointer_position;
-            this.move_pointer_on_switch = c.move_pointer_on_switch;
         } else {
             log(`error loading conf: ${conf.why}`)
         }
--- a/src/prefs.ts
+++ b/src/prefs.ts
@@ -18,6 +18,7 @@ interface AppWidgets {
     snap_to_grid: any,
     window_titles: any,
     show_skip_taskbar: any,
+    move_cursor_on_switch: any,
 }
 
 // @ts-ignore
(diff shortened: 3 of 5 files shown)


mmstick (dismissed):

jacobgkau:
Thank you for implementing this. It seems to be working as I'd expect.

> Does the setting summary `Move cursor to active window when switching via keyboard shortcuts` accurately describe the behaviour here?
> 
> Also looking for suggestions on the label for this setting in the settings window, it doesn't make much sense at the moment.

The GUI currently says `Move Mouse Cursor when Switching`. The word "switching" is a little ambiguous-- it's referring to both navigating windows and switching workspaces. For the longer description, `via keyboard shortcuts` doesn't cover everything since it also takes effect for touchpad gestures (4-finger workspace switching and 3-finger window navigation). I would propose the following:

- `Mouse Cursor Follows Active Window` for the GUI option.
- `Move cursor to active window when navigating with keyboard shortcuts or touchpad gestures` for the longer description.

I also wanted to point out that we still have a hidden option in the `~/.config/pop-shell/config.json` text config called `move_pointer_on_switch`, and the similarity between this new feature and that one could be confusing. That existing option, added in https://github.com/pop-os/shell/pull/646, was specifically for switching between applications using Super+Tab and windows using Alt+Tab (not for Super+Arrow-key navigation).

`move_pointer_on_switch` doesn't seem to work anymore. It's not a regression from this change, it's just not working on 22.04 or 20.04 right now-- the cursor doesn't move to the active window when switching with Alt+Tab or Super+Tab on the same workspace, even if that option is set to `true`. On 22.04 (but not 20.04), the mouse cursor does follow the active window when switching between workspaces with Alt+Tab or Super+Tab if this new option is enabled, regardless of what the old option is set to. We'll probably want to get rid of `move_pointer_on_switch`, and make Alt+Tab/Super+Tab on the same workspace follow the new option consistently with everything else. (That doesn't necessarily need to happen in this PR.)

AKK9:
@jacobgkau Thanks for the suggestion and the extra context. I've updated the naming based on your suggestions and updated the variable/function names too for consistency.

It seems that functionality was removed as part of this commit:
https://github.com/pop-os/shell/commit/b68a38c12af9bcd7cf41d4e1110b1547f2e59c5e

As a result it looks like that config item is not longer used. Want me to remove `move_pointer_on_switch` from `config.ts`?

Yeah I think it's best to keep the Alt + Tab part out of this PR. I don't have the knowledge to tackle this.

AKK9:
On a second look, maybe we could amend the `_finish()` functions on the `AppSwitcherPopup`, `WindowSwitcherPopup` and `CyclerPopup` classes in `imports.ui.altTab`?
https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/js/ui/altTab.js#L271

They could perform their original function, then we could set the mouse pointer position? But modifying prototype is probably not a good idea. Given my lack of knowledge here maybe this is best for a separate PR.

jacobgkau:
> It seems that functionality was removed as part of this commit: [b68a38c](https://github.com/pop-os/shell/commit/b68a38c12af9bcd7cf41d4e1110b1547f2e59c5e)
> 
> As a result it looks like that config item is not longer used. Want me to remove `move_pointer_on_switch` from `config.ts`?

I see. Yes, if that config item is not being used anywhere (and I don't see anywhere it's being used), let's remove it so nobody sees it and thinks it's related to what's being added now. (I expect it might stay in the config file on existing installations where the file's already been created, but no reason to keep creating it for new installations if it doesn't do anything.)

We don't have to worry about getting Super+Tab/Meta+Tab pointer movement working right now, the feature the PR is adding is off by default and only part of the advanced settings, and will be useful enough as it is.

jacobgkau (approved):
Regression testing passed, tested with default settings and with the new option enabled:
[shell-pr1417-jammy.txt](https://github.com/pop-os/shell/files/8642204/shell-pr1417-jammy.txt)
[shell-pr1417-jammy-nopointermovement.txt](https://github.com/pop-os/shell/files/8642328/shell-pr1417-jammy-nopointermovement.txt)

Having some problems with focus being lost when moving windows between workspaces, but that is not a regression:
- https://github.com/pop-os/shell/issues/1415 
- https://github.com/pop-os/shell/issues/1421

mmstick (approved):

Issue #1440, opened 2022-05-17
Maximized floating window returns to previous state after losing & regaining focus


weebs:
**(1) Issue/Bug Description:**

When maximizing a window included in the floating windows exception list, it will return to an un-maximized state after switching focus to another window, then switching back

**(2) Steps to reproduce (if you know):**

- Add a window to the floating windows list
- Maximize the window
- Focus another window
- Switch focus back to the original maximized window

**(3) Expected behavior:**

The window should remain maximized

**(4) Distribution (run `cat /etc/os-release`):**
 
NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=jammy
UBUNTU_CODENAME=jammy
LOGO=distributor-logo-pop-os

**(5) Gnome Shell version:**

GNOME 42

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**

pop-shell:
  Installed: 1.1.0~1652463555~22.04~034fb14
  Candidate: 1.1.0~1652463555~22.04~034fb14
  Version table:
 *** 1.1.0~1652463555~22.04~034fb14 1001
       1001 http://apt.pop-os.org/release jammy/main amd64 Packages
       1001 http://apt.pop-os.org/release jammy/main i386 Packages
        100 /var/lib/dpkg/status

**(7) Where was Pop Shell installed from:**

Included as part of PopOS install

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**

Two monitors, one is a built in laptop display (oryx pro), another is a 2560x1440 144Hz HDMI monitor above it

**(9) Other Installed/Enabled Extensions:**

Dash 2 Panel
Sound Input & Output Device Chooser
CPU Power Manager

**(10) Other Notes:**

assert-0:
I have the same issue on 22.04 LTS.
It is also on a dual monitor setup (4k@60Hz and 1440p@144Hz),
and shell version is also `1.1.0~1652463555~22.04~034fb14`.
Also using CPU Power Manager.

(edit)
[...]

PR #1461, opened 2022-06-02
Feature: Maintain maximized windows

### LINKED ISSUE #1458: Maximized IntelliJ window turns non-maximized after opening search files dialog with tile windows activated

With IntelliJ and another app, like Nautilus, opened in a workspace with tile windows activated, if I maximize IntelliJ and open the search file dialog with the shortcut, the Intellij main window turns non-maximized.

It's started to happen a few days ago.

See Issue/Bug Description.

**(3) Expected behavior:**

The IntelliJ main window should stay maximized after opening the search dialog.

22.04 LTS

**(5) Gnome Shell version:**

42.1

```1.1.0~1653580316~22.04~42102e9```

Pop OS 22.04 Clean Install

1 horizontal monitor, 2560 X 1080

None

**(10) Other Notes:**

https://user-images.githubusercontent.com/17600341/170778904-57afb3b3-3298-4f43-bfba-4bb2e4aae2b2.mp4

### PR DESCRIPTION
**Pull request by pkoutsovasilis**
This PR provides a "solution" to [this issue](https://github.com/pop-os/shell/issues/1458). Specifically, it introduces a new option, in the extension settings, which when enabled maximized windows are maintained on focus change events.

For my workflow this feature is really useful. If you find it useful as well, I am more than happy to perform any requested changes.

Cheers

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -244,7 +244,11 @@ export class Ext extends Ecs.System<ExtEvent> {
         this.dbus.Launcher = () => this.window_search.open(this)
 
         this.dbus.WindowFocus = (window: [number, number]) => {
-            this.windows.get(window)?.activate()
+            const target_window = this.windows.get(window)
+            if (target_window) {
+                target_window.activate()
+                this.on_focused(target_window)
+            }
             this.window_search.close()
         }
 
@@ -628,7 +632,6 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     focus_left() {
-        this.unmaximize_workspace(null)
         this.stack_select(
             (id, stack) => id === 0 ? null : stack.tabs[id - 1].entity,
             () => this.activate_window(this.focus_selector.left(this, null))
@@ -636,7 +639,6 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     focus_right() {
-        this.unmaximize_workspace(null)
         this.stack_select(
             (id, stack) => stack.tabs.length > id + 1 ? stack.tabs[id + 1].entity : null,
             () => this.activate_window(this.focus_selector.right(this, null))
@@ -644,12 +646,10 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     focus_down() {
-        this.unmaximize_workspace(null)
         this.activate_window(this.focus_selector.down(this, null))
     }
 
     focus_up() {
-        this.unmaximize_workspace(null)
         this.activate_window(this.focus_selector.up(this, null))
     }
 
@@ -890,7 +890,7 @@ export class Ext extends Ecs.System<ExtEvent> {
             let prev = this.windows.get(this.prev_focused[0]);
             let is_attached = this.auto_tiler.attached.contains(this.prev_focused[0]);
 
-            if (prev && prev !== win && is_attached && prev.actor_exists() && prev.name(this) !== win.name(this)) {
+            if (prev && prev !== win && is_attached && prev.actor_exists() && prev.name(this) !== win.name(this) && prev.workspace_id() === win.workspace_id()) {
                 if (prev.rect().contains(win.rect())) {
                     if (prev.is_maximized()) {
                         prev.meta.unmaximize(Meta.MaximizeFlags.BOTH);
@@ -995,25 +995,24 @@ export class Ext extends Ecs.System<ExtEvent> {
     }
 
     /** Unmaximize any maximized windows on the same workspace. */
-    unmaximize_workspace(win: Window.ShellWindow | null) {
+    unmaximize_workspace(win: Window.ShellWindow) {
         if (this.auto_tiler) {
             let mon
             let work
 
-            if (win == null) {
-                [mon, work] = this.workspace_id()
-            } else {
-                mon = win.meta.get_monitor()
-                work = win.meta.get_workspace().index()
+            if (!win.is_tilable(this)) {
+                return
             }
+            
+            mon = win.meta.get_monitor()
+            work = win.meta.get_workspace().index()
 
             for (const [, compare] of this.windows.iter()) {
                 const is_same_space = compare.meta.get_monitor() === mon
-                    && compare.meta.get_workspace().index() === work;
-                if (is_same_space && !this.contains_tag(compare.entity, Tags.Floating)) {
-                    if (compare.is_maximized()) {
+                    && compare.meta.get_workspace().index() === work ;
+
+                if (is_same_space && !this.contains_tag(compare.entity, Tags.Floating) && compare.is_maximized() && win.entity[0] !== compare.entity[0]) {
                         compare.meta.unmaximize(Meta.MaximizeFlags.BOTH);
-                    }
                 }
             }
         }
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ debian/*
 !debian/*install
 .confirm_shortcut_change
 .vscode
+node_modules

### CODE REVIEW

**Comment by mmstick** [comment]
Wouldn't this be fixed by avoiding the unmaximize on focus of a floating window? That is the intended fix for this.

**Comment by pkoutsovasilis** [comment]
@mmstick what you mention indeed covers the case of floating windows, but what if I don't have a floating window and just need a normal window for some reason maximized (I think this is captured by [this comment](https://github.com/pop-os/shell/issues/1458#issuecomment-1142055938)). This PR maintains maximized any window type, but as soon as you alt-tab to a windows "underneath" or open a new window on the same monitor/workspace, everything goes back to normal which makes a lot of sense to me (kinda like maximize => zoom-in with tiling => auto zoom-out feature). Again I totally get that this behavior maybe is not aligned with what the pop-os shell is trying to achieve/offer.

**Review by mmstick — DISMISSED** [review]
I'd still like to reach a better default behavior though

**Comment by pkoutsovasilis** [comment]
@mmstick by default this behavior I describe above is disabled, so no changes in the default/average user experience. That said, I am more than happy to help if I can with what you think it would be more appropriate

**Comment by jacobgkau** [comment]
> what if I don't have a floating window and just need a normal window for some reason maximized (I think this is captured by [this comment](https://github.com/pop-os/shell/issues/1458#issuecomment-1142055938)).

That comment is about wanting maximization retained on workspace switch, which is also something that might make sense as default behavior.

> This PR maintains maximized any window type, but as soon as you alt-tab to a windows "underneath" or open a new window on the same monitor/workspace, everything goes back to normal

This sounds like the intended behavior of maximization. Besides handling floating window focus and workspace switch, what else normally doesn't work that this would solve? (I haven't tried this yet, but it's building on our build server for testing.)

**Comment by jacobgkau** [comment]
Okay, testing this now, it does look like the new setting is fixing maximization behavior, particularly for workspace switching. Windows are still unmaximized when changing focus on the same workspace via keyboard or mouse. I'm not opposed to having options for changing how maximization is handled, but it seems like this fix could simply be applied to the default behavior without adding any options in this case, unless I'm missing something else so far.

**Comment by mmstick** [comment]
If this seems like a better default then we can make it so.

**Comment by pkoutsovasilis** [comment]
@jacobgkau I am gonna be honest and say that it did not cross my mind that this was the intended behavior of maximized windows... And in my effort not to step on anybodys toes I made this behavior tunable :smile: Mainly it addresses the two comments highlighted by the aforementioned issue, with an extra twist that when you alt-tab (or switch via launcher) in an "underneath" window or open a new window on the same monitor and workspace the window is given back to auto-tiling and all this magic happens.. I can provide some visual feedback if necessary

**Comment by mmstick** [comment]
We can remove the option and make it the default behavior.

**Comment by jacobgkau** [comment]
The intent of unmaximizing on focus change is to not have tiled windows "floating" over a maximized window (with an unseen gap in the tree) and more generally to not have a maximized window break keyboard navigation. Changing workspaces shouldn't move or resize anything-- a user expects a workspace to look the same when coming back to it as it looked before switching away.

> Mainly it addresses the two comments highlighted by the aforementioned issue, with an extra twist that when you alt-tab (or switch via launcher) in an "underneath" window or open a new window on the same monitor and workspace the window is given back to auto-tiling and all this magic happens.

Alt-tab on the same workspace already unmaximized a window and I'm seeing it continue to do so now (whether the new setting is enabled or disabled.) Alt-tab between workspaces used to unmaximize and no longer does with the setting enabled, which is expected as described above.

Focusing a window with the launcher used to leave the maximized window maximized, and still does, whether the setting is enabled or disabled. I think I would expect this to unmaximize the window (to be consistent with alt-tab and clicking windows in the Workspaces menu.)

**Comment by pkoutsovasilis** [comment]
> Focusing a window with the launcher used to leave the maximized window maximized, and still does, whether the setting is enabled or disabled. I think I would expect this to unmaximize the window (to be consistent with alt-tab and clicking windows in the Workspaces menu.)

@jacobgkau thanks for the feedback! definitely I missed something, I will have a look and push the commit :smile:

**Comment by pkoutsovasilis** [comment]
@jacobgkau switching to a window from launcher now unmaximizes any window on the same monitor and workspace

**Comment by pkoutsovasilis** [comment]
@mmstick @jacobgkau in order to support treating as ```floating``` the window of the intellij search dialog (I assume this applies to other jetbrains products as well) I extended the floating rules and introduced the field of ```window_type```. Initially, I tried to nail this down only by the existing criteria of ```wm_class``` and the ```wm_title``` but unfortunately when the main window of intellij starts it has the same wm_class and wm_title as the search dialog. So, now I am facing the issue that the ```Floating Windows Expeptions``` (from tray panel) is not opening and I think it has something to do with the traspile and the ```Meta``` import. At this point, I am wondering shouldn't all dialog windows treated as floating ones? If no, could you give me some feedback on the transpile obstacle? thanks

**Comment by mmstick** [comment]
The dialog window type should be floating regardless of exceptions. Is this a case where the window type is changing after the window is opened? Because only normal window types are permitted to tile.

**Comment by pkoutsovasilis** [comment]
Ok so this is as I suspected then, dialog window type should be floating. To this end, shouldn't the ```window_shall_float(...)``` check if the window type is a dialog and return true? because now it does not. In the case of intellij, the sequence is the following, spashscreen (normal type) -> welcome to Intellij or already opened project (normal type) -> [ctrl+shift+f] brings up a new global search window (dialog type).

**Comment by mmstick** [comment]
That is checked here: https://github.com/pop-os/shell/blob/master_jammy/src/window.ts#L312

**Comment by pkoutsovasilis** [comment]
thanks for that pointer @mmstick, I definitely missed this out. Now this simplifies my approach by a lot! You see the ```is_floating``` caught my eye and tried to achieve this from that way :smile: but I was wrong. With your help, now focused floating windows wont unmaximize a previously maximized window

File: schemas/org.gnome.shell.extensions.pop-shell.gschema.xml [hunk 1]
```diff
@@ -22,6 +22,11 @@
             <summary>Show title bars on windows with server-side decorations</summary>
         </key>
 
+        <key type="b" name="maintain-maximized-windows">
+            <default>false</default>
+            <summary>Maintain maximized windows on changing focused window</summary>
+        </key>
```
mmstick: Needs to be removed

src/extension.ts
```diff
@@ -336,6 +340,9 @@ export class Ext extends Ecs.System<ExtEvent> {
 
                     case WindowEvent.Size:
                         if (this.auto_tiler && !win.is_maximized() && !win.meta.is_fullscreen()) {
+                            if (this.contains_tag(win.entity, Tags.MaintainMaximized)) {
+                                this.delete_tag(win.entity, Tags.MaintainMaximized)
+                            }
```
mmstick: Same here

src/extension.ts
```diff
@@ -1895,6 +1944,9 @@ export class Ext extends Ecs.System<ExtEvent> {
                 case 'show-title':
                     this.on_show_window_titles();
                     break;
+                case 'maintain-maximized-windows':
+                    this.on_maintain_maximized_widows();
+                    break;
```
mmstick: Here

src/extension.ts
```diff
@@ -1557,6 +1582,10 @@ export class Ext extends Ecs.System<ExtEvent> {
 
     /** Handle window minimization notifications */
     on_minimize(win: Window.ShellWindow) {
+        if (this.contains_tag(win.entity, Tags.MaintainMaximized)) {
+            this.delete_tag(win.entity, Tags.MaintainMaximized)
+        }
```
mmstick: Here

src/extension.ts
```diff
@@ -1529,6 +1550,10 @@ export class Ext extends Ecs.System<ExtEvent> {
 
     /** Handle window maximization notifications */
     on_maximize(win: Window.ShellWindow) {
+        if (this.settings.maintain_maximized_windows()) {
+            this.add_tag(win.entity, Tags.MaintainMaximized)
+        }
+
```
mmstick: Here

src/extension.ts
```diff
@@ -1662,6 +1691,26 @@ export class Ext extends Ecs.System<ExtEvent> {
...
+        for (const window of this.windows.values()) {
+            if (maintain_maximized_windows) {
+                if (window.is_maximized() && !this.contains_tag(window.entity, Tags.MaintainMaximized)) {
+                    this.add_tag(window.entity, Tags.MaintainMaximized);
+                }
+            } else {
+                if (this.contains_tag(window.entity, Tags.MaintainMaximized)) {
+                    this.delete_tag(window.entity, Tags.MaintainMaximized);
+                }
```
mmstick: These

src/prefs.ts
```diff
@@ -35,7 +36,12 @@ function settings_dialog_new(): Gtk.Container {
         Settings.sync();
     });
 
-    app.snap_to_grid.set_active(ext.snap_to_grid());
+    app.maintain_maximized_windows.set_active(ext.maintain_maximized_windows());
+    app.maintain_maximized_windows.connect('state-set', (_widget: any, state: boolean) => {
+        ext.set_maintain_maximized_windows(state);
+        Settings.sync();
+    });
```
mmstick: This

src/extension.ts
```diff
@@ -1011,8 +1025,15 @@ export class Ext extends Ecs.System<ExtEvent> {
...
-                        compare.meta.unmaximize(Meta.MaximizeFlags.BOTH);
+                    if (win !== null ) {
+                        if (compare.is_maximized() && win.entity[0] !== compare.entity[0]) {
+                            compare.meta.unmaximize(Meta.MaximizeFlags.BOTH);
+                            this.delete_tag(compare.entity, Tags.MaintainMaximized)
+                        }
+                    } else {
+                        if (compare.is_maximized() && !this.contains_tag(compare.entity, Tags.MaintainMaximized)) {
+                            compare.meta.unmaximize(Meta.MaximizeFlags.BOTH);
+                        }
```
mmstick: This

src/prefs.ts
```diff
@@ -17,6 +17,7 @@ interface AppWidgets {
     smart_gaps: any,
     snap_to_grid: any,
     window_titles: any,
+    maintain_maximized_windows: any,
```
mmstick: This

src/prefs.ts
```diff
@@ -115,12 +121,18 @@ function settings_dialog_view(): [AppWidgets, Gtk.Container] {
         label: "Mouse Cursor Follows Active Window",
         xalign: 0.0
     });
+
+    let maintain_maximized_windows_label = new Gtk.Label({
+        label: "Maintain Maximized Windows",
+        xalign: 0.0
+    });
```
mmstick: This

src/settings.ts
```diff
@@ -54,6 +54,7 @@ const GAP_INNER = "gap-inner";
 const GAP_OUTER = "gap-outer";
 const ROW_SIZE = "row-size";
 const SHOW_TITLE = "show-title";
+const MAINTAIN_MAXIMIZED_WIDOWS = "maintain-maximized-windows";
```
mmstick: These

src/tags.ts
```diff
@@ -1,4 +1,5 @@
 export var Tiled = 0;
 export var Floating = 1;
 export var Blocked = 2;
-export var ForceTile = 3;
\ No newline at end of file
+export var ForceTile = 3;
+export var MaintainMaximized = 4;
```
mmstick: This

mmstick (changes requested):

pkoutsovasilis:
@mmstick I saw the requested changes and I guess that we are going to make this the default behavior. But if we do so, we are actually making ```unmaximize_workspace(null)``` obsolete. If that's the case, should I remove the aforementioned calls to ```unmaximize_workspace``` with null?

mmstick:
Sure

pkoutsovasilis:
@mmstick this now alters the default behavior of maximized windows as described in the comments above

mmstick (approved):

jacobgkau (approved):
Fixes both issues described in https://github.com/pop-os/shell/issues/1458 (IntelliJ no longer unmaximizes when opening the search dialogue with double-pressing Shift, and switching workspaces no longer causes a maximized window to unmaximize.)

Regression testing passed: [shell-pr1461-jammy.txt](https://github.com/pop-os/shell/files/8847798/shell-pr1461-jammy.txt)

frezbo:
this should also fix this https://github.com/pop-os/shell/issues/1440#issuecomment-1145700475= :+1:

PR #1468, opened 2022-06-07
feat: Add fullscreen launcher pref

### LINKED ISSUE #1465: Launcher does not appear when a full screen window is active
I've noticed in the last week or so the pop launcher refuses to appear when I hit the Super key and a full screen window is active.

I can reproduce it in a multiple places, namely when I'm using Remmina to RDP into my work machine or watching Netflix full screen.

Not sure if this is a bug or an intentional change of behavior, but it's definitely less convienent as now I have to minimize what I'm doing in my full screen window if I want to invoke another application via the launcher.

mmstick:
Adds an option for those who want to enable it

Closes #1465

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -7,6 +7,11 @@
             <summary>Show a hint around the active window</summary>
         </key>
 
+        <key type="b" name="fullscreen-launcher">
+            <default>false</default>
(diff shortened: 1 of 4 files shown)


n3m0-22 (dismissed):
Tested with a few fullscreen apps and games. This works well, and provides a nice solution for both use cases.

jackpot51 (approved):

n3m0-22 (approved):
All regression testing passed with the new setting both on and off.

[shell-testing.txt](https://github.com/pop-os/shell/files/8855020/shell-testing.txt)

Issue #1467, opened 2022-06-06
Windows Unmaximized on Screen Lock/Unlock


jacobgkau:
**(1) Issue/Bug Description:** If a window is maximized when the screen is locked, it is unmaximized when the screen is unlocked.

**(2) Steps to reproduce (if you know):**
1. Enable tiling with several windows open.
2. Maximize a window.
3. Lock and unlock the screen.

**(3) Expected behavior:** If a window was maximized before the screen was locked, it should still be maximized after the screen is unlocked.

**(4) Distribution (run `cat /etc/os-release`):** Pop!_OS 22.04 LTS

**(5) Gnome Shell version:** 42

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):** 44ca2cd (`master`) and 2eebb41 (`testing-shell-pr1461`)

**(7) Where was Pop Shell installed from:** Pop!_OS release and staging repos.

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):** 1x 1080p

**(9) Other Installed/Enabled Extensions:** Pop!_OS defaults only.

**(10) Other Notes:** Not caused by or fixed by https://github.com/pop-os/shell/pull/1461.

PR #1469, opened 2022-06-07
fix: update workareas only after screen unlock

pkoutsovasilis:
This PR offers a solution to [this issue](https://github.com/pop-os/shell/issues/1467).

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2333,6 +2333,14 @@ export class Ext extends Ecs.System<ExtEvent> {
             return
         }
 
+        if (workareas_only) {
+            if (this.workareas_update !== null) {
+                GLib.source_remove(this.workareas_update)
+                this.workareas_update = null
+            }
+            return;
+        }
+
         // Update every tree on each display with the new dimensions
         const update_tiling = () => {
             if (!this.auto_tiler) return


mmstick (approved):

jacobgkau (changes requested):
This fixes https://github.com/pop-os/shell/issues/1467, but after the change, if I open up windows and then restart GNOME Shell with Alt-F2, r, Enter, then the tree goes behind the top bar and dock:

Screenshot from 2022-06-08 10-00-47

Restarting GNOME Shell is somewhat common in both testing and troubleshooting, so that seems like an issue. If I close _all_ windows and then open some windows back up, that doesn't happen anymore; but then if I e.g. move the dock, then the tree doesn't update to get out of the way of that either.

Screenshot from 2022-06-08 10-03-28

pkoutsovasilis:
good catch @jacobgkau I admit that I didn't occur to me to test this on gnome reset. I suspect that on gnome reset the param ```workareas_only``` is "erroneously" set to ```true``` where it should be ```false```... I will come back on this

pkoutsovasilis:
this is no longer relevant 🙂

Issue #1287, opened 2021-12-22
Changing tiling orientation by dragging is too easy


expe:
**(1) Issue/Bug Description:**

It can be easy to inadvertently change the tiling layout by accidentally dragging windows. See video below of where I drag the window a tiny distance down, but it triggers an orientation change. The change is expected, but perhaps there should be a minimum drag distance to trigger it?

https://user-images.githubusercontent.com/4994123/147038557-a0dcb133-c020-4f47-902a-583ad0a1a1eb.mp4

**(2) Steps to reproduce (if you know):**

- Open 2 windows (windows should be tiled side by side)
- Drag one of the windows down (1 pixel might be enough)
- The orientation will change to top and bottom

**(3) Expected behavior:**

Orientation should not change if the drag distance is small.

**(4) Distribution (run `cat /etc/os-release`):**
 ```sh
NAME="Pop!_OS"
VERSION="21.10"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 21.10"
VERSION_ID="21.10"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=impish
UBUNTU_CODENAME=impish
LOGO=distributor-logo-pop-os
```

**(5) Gnome Shell version:**
`40.5-1ubuntu2pop0~1633630918~21.10~91d3049`

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**
```sh
pop-shell:
  Installed: 1.1.0~1639174177~21.10~de901ae
  Candidate: 1.1.0~1639174177~21.10~de901ae
  Version table:
 *** 1.1.0~1639174177~21.10~de901ae 1001
       1001 http://apt.pop-os.org/release impish/main amd64 Packages
       1001 http://apt.pop-os.org/release impish/main i386 Packages
        100 /var/lib/dpkg/status
```

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**

3440 x 1440

PR #1482, opened 2022-06-18
feat: Ignore accidental window movements/grab ops

michaelmob:
## Problem

*Also mentioned in: #1287*

A common **problem** I have is messing up my window tree by accidentally grabbing the titlebar of a window or by [holding the super key](https://wiki.gnome.org/Gnome3CheatSheet#Resizing_Windows) and accidentally clicking anywhere on the window, dragging it. These accidental grabs/movements/re-positions of the window cause an instant reflow of the tree that can be disorienting and distracting focus from your main task.

https://user-images.githubusercontent.com/4530010/174450402-b7be9b72-35d0-4824-a7b3-fcea9bb9cabb.mp4

* * *

## Solution

My preferred **solution** is to allow a small duration of time for the extension to ignore a window grab op/movement.

When we start a grab op, we set a current timestamp in the new `grab_time` property. When the grab op ends, that stored value is subtracted from the new current timestamp. If the result in milliseconds is less than the specified `min_drag_ms` then the `op` is set to `null` so the tree reflows back into position without being changed, essentially ignoring the window grab op/movement.

I don't know about you, but I know I'm not making precise window movement decisions in under 100ms. 100ms has worked so far for me and it could probably be as high as 200ms. I think 100ms is a good duration of buffer time for the autotiler to ignore an accident.

https://user-images.githubusercontent.com/4530010/174450413-3caded1a-5512-4459-a9c6-642cf896e2a4.mp4

* * *

### Another lesser solution (not implemented)

Another solution is to compare the before and after window positions and determine if it has moved a far enough distance to warrant a tree change. This method would require a lot more logic. I consider this solution to be inferior to the paragraph above as High DPI mice/high resolution monitors would thwart this measure and we'd probably have to use percentages instead of pixels.

--

*Work still in-progress?* I haven't thoroughly tested this on any machine other than my own with GNOME 42.1. I'm very open to conversation and suggestions on the mechanics of this thing.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -127,6 +127,9 @@ export class Ext extends Ecs.System<ExtEvent> {
 
     drag_signal: null | SignalID = null
 
+    /** Minimum amount of time for a window to be dragged to apply updates */
+    min_drag_ms: number = 100;
+
     /** If set, the user is currently selecting a window to add to floating exceptions */
     exception_selecting: boolean = false;
 
@@ -147,6 +150,9 @@ export class Ext extends Ecs.System<ExtEvent> {
 
     /** Information about a current possible grab operation */
     grab_op: GrabOp.GrabOp | null = null;
+
+    /** Timestamp of most recent window grab by mouse */
+    private grab_time: number = 0;
 
     /** A display config update is triggered on a workspace addition */
     ignore_display_update: boolean = false;
@@ -1065,6 +1071,11 @@ export class Ext extends Ecs.System<ExtEvent> {
             return;
         }
 
+        // Prevent accidental mouse drag
+        if ((+new Date()) - this.grab_time < this.min_drag_ms) {
+            op = undefined;  // reflow without updating tree
+        }
+
         this.on_grab_end_(win, op);
         this.unset_grab_op()
     }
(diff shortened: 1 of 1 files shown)


src/extension.ts
```diff
@@ -1396,6 +1407,7 @@ export class Ext extends Ecs.System<ExtEvent> {
     /** Triggered when a grab operation has been started */
     on_grab_start(meta: null | Meta.Window, op: any) {
         if (!meta) return
+        this.grab_time = +new Date();  // timestamp
```
michaelmob, line 1410: Is this an appropriate way to get a timestamp in gjs?

I wanted to use `this.grab_time = global.get_current_time();` but it only seems to return `0`. Maybe it doesn't do what I think it does, couldn't find docs.

jacobgkau:
A couple of observations:

- This does prevent some extremely quick movements from taking effect (that used to take effect before the change.)
- However, there are still some movements that are quick enough for the overlay to not show up, but slow enough to take effect.

https://user-images.githubusercontent.com/7199422/174653032-99432d96-e272-4e39-818e-8bf3979e6e5b.mp4

The delay between click/drag and the overlay showing up is already present in the currently-released version and is not a regression. This change has me wondering if the overlay should show up as soon as a click/drag would actually take effect. While 100ms is usually quicker than an intentional move, I can create false positives by e.g. flinging a window into a corner (especially with a non-touchpad mouse), and the fact that there's no visual feedback for when I've held down the button long enough could be considered confusing.

PR #1486, opened 2022-06-21
feat!: Rework resize operations with resize mode

mmstick:
Pressing Super + R will now enter a resize mode, which has a visual indicator to show whether the resizes will grow or shrink the window. Arrow keys and hjkl will grow the window in the four directions by default. Holding Shift will invert the arrows and allow shrinking the window from the four sides.

### DIFF
--- a/dark.css
+++ b/dark.css
@@ -55,3 +55,9 @@
 .pop-shell-tab-urgent {
     background: #D00;
 }
+
+.pop-shell-resize-hint {
+    background: #EDEDED;
+    padding: 12px;
+    border-radius: 32px;
+}
--- a/highcontrast.css
+++ b/highcontrast.css
@@ -59,3 +59,9 @@
 .pop-shell-entry:indeterminate {
     font-style: italic
 }
+
+.pop-shell-resize-hint {
+    background: #EDEDED;
+    padding: 12px;
+    border-radius: 32px;
+}
--- a/keybindings/10-pop-shell-move.xml
+++ b/keybindings/10-pop-shell-move.xml
@@ -17,13 +17,10 @@
   <KeyListEntry name="pop-monitor-right" description="Move window to rightward monitor"/>
   <KeyListEntry name="pop-monitor-up" description="Move window to upper monitor"/>
   <KeyListEntry name="tile-reject" description="Cancel changes"/>
-  <KeyListEntry name="tile-resize-left" description="Resize window smaller"/>
-  <KeyListEntry name="tile-resize-right" description="Resize window larger"/>
-  <KeyListEntry name="tile-resize-up" description="Resize window shorter"/>
-  <KeyListEntry name="tile-resize-down" description="Resize window taller"/>
   <KeyListEntry name="tile-swap-down" description="Swap window down"/>
   <KeyListEntry name="tile-swap-left" description="Swap window left"/>
   <KeyListEntry name="tile-swap-right" description="Swap window right"/>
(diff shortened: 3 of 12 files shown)


jacobgkau (changes requested):
A couple of problems on 1083bdd:

- The Super-Enter shortcut still works to enter window management mode, but doesn't allow resizing windows anymore; with no visual indication, this is going to be confusing to current users upgrading.
- With two windows split horizontally on a 1080p display, I can can only make a window larger without holding shift, and I can only make a window smaller while holding shift; the arrow indicators seem to indicate that I should be able to use the left or right arrow key in either mode (with the only difference being inverted while holding shift.)
    - Upon reading your description again, I understand the reason this isn't working is because I can't grow a window against the edge of a display; perhaps the on-screen arrow indicators should only be present for directions where resizing is possible, or else maybe the control scheme could use further consideration.

Separately from these issues, a change like this is going to need @pop-os/ux input/approval.

jacobgkau:
For reference, here is what the indicator currently looks like for resizing, without and with the shift key:

mpv-shot0076
mpv-shot0077

I noticed a couple of other issues while getting those pictures:

- The Print Screen key does not work while the resize mode is open. (It works in the current window management mode.)
- Clicking anywhere on the screen while the resize mode is active does nothing; current window management mode does not exit on click (perhaps it should), but it doesn't entirely block the mouse like this.
- Because of the background color chosen for the resize indicator, there's a lack of contrast with Terminal (and probably other apps using Pop colors):
mpv-shot0078

jacobgkau:
The new arrow placement is easier to understand:

https://user-images.githubusercontent.com/7199422/183953855-33db7c6d-6d48-48a8-8ead-d0272f1300eb.mp4

It does seem strange that I need to e.g. use Shift in order to be able to use two of the arrow keys, and release Shift to be able to use the other two, in the above example. It makes me wonder why I need shift at all. Of course, the reason is that sometimes a window has three or more resizeable edges:

https://user-images.githubusercontent.com/7199422/183953872-d8eb4513-7131-4fef-a3b2-9837946d8f41.mp4

But still feels a little strange.

These points still stand:

- The Super-Enter shortcut still works to enter window management mode, but doesn't allow resizing windows anymore; with no visual indication, this is going to be confusing to current users upgrading.
- The Print Screen key does not work while the resize mode is open. (It works in the current window management mode.)

mmstick:
Waiting for new designs from UX for these issues

siddhpant:
Any update on this one? Seems like a very helpful feature as currently I'm clueless what would happen when resizing.

For example, on two tiles, for the right pane, pressing ctrl+right increases to the left and ctrl+left decreases to the right. Shouldn't it be the reverse? Maybe this is a bug, but if this fixes it then I don't want to open a new issue.

jacobgkau:
This has been implemented in [COSMIC Comp](https://github.com/pop-os/cosmic-comp), part of the new COSMIC Epoch desktop environment written in Rust. Our GNOME session is essentially in maintenance mode, so it's unlikely a large UX change like this will receive additional development resources.

> For example, on two tiles, for the right pane, pressing ctrl+right increases to the left and ctrl+left decreases to the right.

The current behavior defaults to Shift + Arrows in management mode for resizing, not Ctrl. That said, with the current Pop!_Shell UX, this isn't technically a bug. Shift-Right increases the window size because right == increase, and the only possible direction to do that would be to the left. This can, of course, be unintuitive in some circumstances, which is why we switched to the UX presented here for COSMIC Epoch.

PR #1499, opened 2022-07-25
feat: global tiling

### LINKED ISSUE #1400: Can't move tiles without entering adjustment mode
In i3 super + shift + direction moves the tile within a workspace, in and out of a stack, then to another monitor. It allows for seamless selection of tiles with super + direction and movement by also holding shift. Please allow for a setting where super + shift + direction behaves like i3 and we don't need to waste time entering and exiting adjustment mode, this window manager is almost perfect. Thanks!

sevenautumns:
A (somewhat) hacky solution for #1400.
Introducing 4 new dconf names:
- tile-move-down-global
- tile-move-left-global
- tile-move-right-global
- tile-move-up-global

which are per-default empty.
These functions are for moving tiles outside of the management mode for a (optional) more i3-like experience.

For implementing this feature, I hijacked `move`, `move_up`,  `move_left`, `move_right`, `move_down` and requested that an additional argument (the moved window) is passed. By doing this, a calling function can either specify `this.window` if its called from within the management mode or `ext.focus_window()?.entity` if its called from a global state.

I feel like because of this additional window-entity parameter, all of the move functions should be refactored into the AutoTiler. then the tiler could use the AutoTiler for moving windows (like its currently done with the `toggle_stacking`-function)

### DIFF
--- a/keybindings/10-pop-shell-move.xml
+++ b/keybindings/10-pop-shell-move.xml
@@ -6,6 +6,10 @@
   <KeyListEntry name="tile-move-left" description="Move window left"/>
   <KeyListEntry name="tile-move-right" description="Move window right"/>
   <KeyListEntry name="tile-move-up" description="Move window up"/>
+  <KeyListEntry name="tile-move-down-global" description="Move window down outside management mode"/>
+  <KeyListEntry name="tile-move-left-global" description="Move window left outside management mode"/>
+  <KeyListEntry name="tile-move-right-global" description="Move window right outside management mode"/>
+  <KeyListEntry name="tile-move-up-global" description="Move window up outside management mode"/>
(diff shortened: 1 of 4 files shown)


mmstick (approved):

jacobgkau (approved):
Regression testing passed: [shell-pr1499-jammy.txt](https://github.com/pop-os/shell/files/9184255/shell-pr1499-jammy.txt)

The new shortcuts are unassigned by default, so this does not cause any change of default behavior.

The new shortcuts work as expected after manually configuring keybindings for them. I tested with Super + Alt + direction keys, Super + WASD, and overwriting the default Super + Shift + direction keys.

I had some trouble getting keybinding changes to apply and reset, but that appears to be an unrelated issue with GNOME Control Center handling keyboard shortcuts (I recreated it on a separate machine without this branch and it's not specific to pop-shell shortcuts.)

PR #1509, opened 2022-08-18
feat: active-hint-border-radius

bennypowers:
## Why?
Better integration with stock GNOME (Fedora, in this case)

## What?
Adds an `active-hint-border-radius` configuration setting to the database and to the panel UI

Screenshot from 2022-08-18 14-47-08

## Notes

This is my first GTK/GNOME rodeo. Pop shell is :man_cook: :kiss:  though, great work!

Reviewers, please take a look at the settings schema and code, as I was not yet able to view the setting with description in dconf-editor

Screenshot from 2022-08-18 14-49-44

### DIFF
--- a/dark.css
+++ b/dark.css
@@ -1,7 +1,7 @@
 .pop-shell-active-hint {
     border-style: solid;
     border-color: #FBB86C;
-    border-radius: 5px;
+    border-radius: var(--active-hint-border-radius, 5px);
     box-shadow: inset 0 0 0 1px rgba(24, 23, 23, 0)
 }
 
--- a/highcontrast.css
+++ b/highcontrast.css
@@ -1,7 +1,7 @@
 .pop-shell-active-hint {
     border-style: solid;
     border-color: #FBB86C;
-    border-radius: 5px;
+    border-radius: var(--active-hint-border-radius, 5px);
     box-shadow: inset 0 0 0 1px rgba(24, 23, 23, 0)
 }
 
--- a/light.css
+++ b/light.css
@@ -1,7 +1,7 @@
 .pop-shell-active-hint {
     border-style: solid;
     border-color: #FFAD00;
-    border-radius: 5px;
+    border-radius: var(--active-hint-border-radius, 5px);
     box-shadow: inset 0 0 0 1px rgba(200, 200, 200, 0);
 }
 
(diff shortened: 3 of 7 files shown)


n3m0-22 (approved):
Tested on Pop and Fedora all testing passed

[shell-testing.txt](https://github.com/pop-os/shell/files/9376339/shell-testing.txt)

The `no schema found` issue is due to it being installed at user level vs system. @bennypowers If you would like it to show in `dconf-editor` the following will fix it for your local install.

`sudo cp schemas/org.gnome.shell.extensions.pop-shell.gschema.xml /usr/share/glib-2.0/schemas/`
`sudo glib-compile-schemas /usr/share/glib-2.0/schemas/`

You may need to logout, reboot, or restart GNOME to see the change.

Example in Fedora:

Screenshot from 2022-08-18 11-53-41

bennypowers:
Amazing, I'll give that a try, and thanks for the review.

I noticed that window content can overflow the border. That may be an issue for a subsequent MR though. See e.g. the corners of the wezterm instance in the above screenshot.

n3m0-22 (changes requested):
That could be an issue. I will need to look into that further.

n3m0-22:
@bennypowers I don't think it is too much of an issue as it doesn't have much impact on window contents unless the radius is set to greater that 30 and/or the window is very small. It seems there is a sweet spot between around 12 - 15 that works well for vanilla gnome. Would you be able to add an upper limit of 15? No lower limit is needed, and the default of 5 is good as it keeps the look consistent for Pop!_OS.

bennypowers:
Initially I tried implementing a [QuickSlider](https://github.com/GNOME/gnome-shell/blob/b7d738dfffab07f84649b647ea9ce639f37079a4/js/ui/quickSettings.js#L149), but that fight was, alas, beyond my experience/interest. So instead I added some optional params to `clamp` and an override to `number_entry`. I set the range to `min: 5, max: 30: value: 5`. I found myself that a radius of 17 looked the best on my desktop

n3m0-22 (approved):
I think this is a good compromise. It allows for enough variation to account for user preference. I suppose a user can still overdo it with the radius, but it is very easy to revert to a more acceptable setting.

Tested again on Pop and Fedora. Testing is still passing. This looks good to me.

Thank you for contributing to Pop!_OS.

mmstick (approved):

PR #1511, opened 2022-08-23
chore(exceptions): Add JavaEmbeddedFrame

sitiom:
Float tray icon of Java applications

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -34,6 +34,7 @@ export const DEFAULT_FLOAT_RULES: Array<FloatRule> = [
     { class: "Gnome-terminal", title: "Preferences – General" },
     { class: "Guake" },
     { class: "Io.elementary.sideload" },
+    { title: "JavaEmbeddedFrame" },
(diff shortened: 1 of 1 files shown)


jacobgkau (approved):
Confirmed this works using JDownloader 2 as an example app. It's strange that the tray icon for apps using this framework show up as a black square in the actual tray and then float their actual icons over the desktop separately, but floating it makes more sense than allowing it to be tiled (it's tiny, non-resizeable, and not able to be interacted with.)

jacobgkau:
Before:
Screenshot from 2022-08-24 14-49-07

After:
Screenshot from 2022-08-24 14-48-02

sitiom:
> It's strange that the tray icon for apps using this framework show up as a black square in the actual tray and then float their actual icons over the desktop separately

It doesn't seem to be the case with Wayland however. The icon is properly displayed in the tray. Example with Megabasterd:
image

13r0ck (approved):
Thanks!

PR #1514, opened 2022-08-31
fix: Allow installation on GNOME 43

carlwgeorge:
I was able to set `org.gnome.shell.disable-extension-version-validation` to true and install the extension on GNOME shell 43 beta.  Everything appears to work as expected.


mmstick (approved):

n3m0-22 (approved):
Tested on Fedora 37 with GNOME 43.beta. All testing that was possible passed. Looks good. 

[gnome-43-pop-shell.txt](https://github.com/pop-os/shell/files/9472099/gnome-43-pop-shell.txt)

PR #1523, opened 2022-09-27
active border radius minimum zero

### LINKED ISSUE #1522: [Feature Request] Allow border radius to be reduced to 0
https://github.com/pop-os/shell/blob/master_jammy/src/panel_settings.ts#L72

Any chance, to set the min setting to zero(0)?

n3m0-22:
This changes the lower bound for the Active Border Radius to zero per user request https://github.com/pop-os/shell/issues/1522. This does not change the default value.

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -9,7 +9,7 @@
 
         <key type="u" name="active-hint-border-radius">
             <default>5</default>
-            <range min="5" max="30"/>
+            <range min="0" max="30"/>
             <summary>Border radius for active window hint, in pixels</summary>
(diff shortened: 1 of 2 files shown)


13r0ck (dismissed):

jacobgkau (changes requested):
This allows me to change the option in the top-right menu below 5 (down to 0), but values below 5 do not take effect or propagate to the gsettings entry. Trying to set the value to below 5 via `gsettings set` also doesn't work. I think there's a remaining limit elsewhere in the code.

Screenshot from 2022-09-28 14-58-55

jacobgkau (approved):
This is working now:
Screenshot from 2022-09-29 09-17-06
Does not affect the default.

mmstick (approved):

PR #1527, opened 2022-10-07
feat: Move mouse focus position config to gnome extension options

zunda-arrow:
This PR moves mouse focus config to the extension options. A problem I had when using Pop shell was figuring out how to make the mouse cursor centered when you select a tiled window. This PR fixes that by putting the configuration option in a more visible place.
`DefaultPointerPosition` is renamed to `FocusPosition` and moved to `src/focus.ts` because it does not belong with the config class after this PR.
If a user set a value in the old config file then updates to this version, the config will be reset. I do not know a good way to respect the old config value.

Is it worth considering merging the "Mouse Cursor Follows Active Window" and "Mouse Cursor Focus Position" options?

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -43,6 +43,11 @@
             <summary>Move cursor to active window when navigating with keyboard shortcuts or touchpad gestures</summary>
         </key>
 
+        <key type="u" name="mouse-focus-location">
+            <default>0</default>
+            <summary>The location the mouse cursor focuses when selecting a window</summary>
+        </key>
+
         <!-- Tiling Options -->
         <key type="u" name="column-size">
             <default>64</default>
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -43,7 +43,7 @@
             <summary>Move cursor to active window when navigating with keyboard shortcuts or touchpad gestures</summary>
         </key>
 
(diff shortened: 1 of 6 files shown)


mmstick:
If it can be done in a backwards-compatible way, sure.

jacobgkau (approved):
The setting in the extension options does work as expected (identically to the old one). As the PR description mentioned, this will effectively reset peoples' setting from the config file; it will also leave the now-removed option in the config file, which could be confusing for people upgrading. On new installations (or for those who manually remove their config file), the option is no longer pre-populated in the config file.

I think this move is justified for a couple of reasons:
- The PR that added this feature, https://github.com/pop-os/shell/pull/646, did not go through UX review, and the option was essentially "hidden" by being in the config file without being exposed through any GUI.
    - Technically, even the options in the Extensions preferences panel are "experimental." Only the settings exposed in the top-right menu (or in the future, Settings -> Desktop) are guaranteed.
- The setting is more logically grouped with the rest of the gsettings than it is with the rest of the config file settings. The gsettings already include e.g. the option to enable or disable cursor movement altogether. The purpose of the config file seems more intended to store less finitely structured data (e.g. the list of manual floating exceptions) that gsettings would not be able to handle easily.
- A postinst script or similar to migrate the setting from the config file to gsettings seems like it may be over-engineered considering the visibility and significance of the setting in question.

`gsettings reset org.gnome.shell.extensions.pop-shell mouse-cursor-follows-location` now works to reset this individual setting to the default state. Regression testing passed: [shell-pr1527-jammy.txt](https://github.com/pop-os/shell/files/9749316/shell-pr1527-jammy.txt)

mmstick (approved):

PR #1551, opened 2022-12-01
feat: add settings toggle for showing stack tab buttons

awisus:
I am personally not a fan of those tab buttons that are presented when windows are stacked.
grafik

The other window tiling manager that I am used to comes without these which is totally fine, too. Take https://ianyh.com/amethyst/ as example.

So let's give users the choice :)

### DIFF
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -33,6 +33,11 @@
             <summary>Show title bars on windows with server-side decorations</summary>
         </key>
 
+        <key type="b" name="show-stack-tab-buttons">
+            <default>true</default>
+            <summary>Show tabs for stacked windows</summary>
+        </key>
+
         <key type="b" name="show-skip-taskbar">
             <default>true</default>
             <summary>Handle minimized to tray windows</summary>
--- a/src/prefs.ts
+++ b/src/prefs.ts
@@ -21,6 +21,7 @@ interface AppWidgets {
     show_skip_taskbar: any,
     smart_gaps: any,
     snap_to_grid: any,
+    stack_tabs_buttons: any,
     window_titles: any,
     mouse_cursor_focus_position: any,
     log_level: any,
@@ -37,6 +38,12 @@ function settings_dialog_new(): Gtk.Container {
     app.window_titles.set_active(ext.show_title());
     app.window_titles.connect('state-set', (_widget: any, state: boolean) => {
         ext.set_show_title(state);
+        Settings.sync();
+    });
+
+    app.stack_tabs_buttons.set_active(ext.show_stack_tab_buttons());
+    app.stack_tabs_buttons.connect('state-set', (_widget: any, state: boolean) => {
+        ext.set_show_stack_tab_buttons(state);
         Settings.sync();
     });
 
@@ -125,6 +132,11 @@ function settings_dialog_view(): [AppWidgets, Gtk.Container] {
         hexpand: true
     })
 
+    const stack_tab_buttons_label = new Gtk.Label({
+        label: "Show tabs for stacked windows",
+        xalign: 0.0
+    })
+
     const snap_label = new Gtk.Label({
         label: "Snap to Grid (Floating Mode)",
         xalign: 0.0
@@ -155,7 +167,7 @@ function settings_dialog_view(): [AppWidgets, Gtk.Container] {
         xalign: 0.0
     })
 
-    const [inner_gap, outer_gap] = gaps_section(grid, 9);
+    const [inner_gap, outer_gap] = gaps_section(grid, 10);
 
     const settings = {
         inner_gap,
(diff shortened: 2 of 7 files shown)


schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
```diff
@@ -33,6 +33,11 @@
             <summary>Show title bars on windows with server-side decorations</summary>
         </key>
 
+        <key type="b" name="show-stack-tab-buttons">
+            <default>true</default>
+            <summary>Allow showing tab buttons for stacked windows</summary>
```
jacobgkau: "Allow showing" is a roundabout way to phrase this. Nothing is being "allowed." The verb should just be "show."

"Tab buttons" is also potentially confusing, since it's unclear if that refers to the tabs themselves or buttons on the tabs (e.g. the close buttons.)

Please simplify this to `Show tabs for stacked windows`.

src/panel_settings.ts
```diff
@@ -48,6 +49,14 @@ export class Indicator {
 
         this.toggle_tiled = tiled(ext)
 
+        this.toggle_stack_tabs_buttons = toggle(
+            _("Show Tab Buttons for Stacked Windows"),
```
jacobgkau: Let's call them `Tabs` instead of `Tab Buttons`, the word "buttons" sounds like it's something on the tabs.

jacobgkau (changes requested):

jacobgkau:
Having stacks without tabs seems very confusing from a UX perspective. There is no way to tell whether a window has had a stack started on it or not; pressing Super-S to start a stack gives no visual feedback. Once the stack is formed with multiple windows, there is no visual indication that multiple windows are present.

https://user-images.githubusercontent.com/7199422/205188323-9672ece1-78cf-4e85-9104-e9001c72eedd.mp4

That said, since this is a non-default option and is in the semi-hidden extension settings (and not the top-right menu), I don't see any reason this shouldn't be included, as long as @mmstick doesn't have any problems with what it's doing to the tab code.

schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
awisus: Ah ok :) Maybe confusion on my side. I understood the Tab Interface to hold both the app's window alongside the button at the top. Therefore I "over"specified it in this way :)

awisus:
@jacobgkau @mmstick should be resolved :)

awisus:
@jacobgkau should be resolved :)

jacobgkau (changes requested):
On f615a45, there are a few issues:

- The option has been added to the top-right menu. This option has not been approved by UX to go there. Given the nature of the option, I think it's best if we keep it only in the advanced settings accessed through the Extensions app; please remove it from the top-right menu.
    Screenshot from 2023-03-01 11-27-21
    - If we did want to run this by UX, the following usability issues would likely need to be addressed (although I still find the interface without tabs confusing and think it's unlikely we'd want to present that as an option in the top-right menu.)
- Enabling/disabling tabs doesn't take effect until focus is moved onto a stack. Ideally, the option would apply to all stacks immediately when it's changed.

https://user-images.githubusercontent.com/7199422/222233800-da5336a4-6475-4aeb-9816-ebd0aa00aa10.mp4

- After tabs have been disabled, the tiler doesn't update to utilize the extra space until windows in or around the stack have been moved. When tabs are disabled, the windows in the stack should be resized to utilize the extra space so there isn't a gap.
    - The tiler also doesn't shrink windows after tabs have been enabled, which breaks the tabs until a manual movement/resizing is performed; when this issue is happening, it's also possible to make the tabs appear under the top bar, which should never be the case.

https://user-images.githubusercontent.com/7199422/222233845-034de382-e165-4b2c-a2ab-0044b4812863.mp4

awisus:
Hi @jacobgkau, thanks for the review. The open issues have been addressed. Please check them out :)

src/prefs.ts
```diff
@@ -125,6 +132,11 @@ function settings_dialog_view(): [AppWidgets, Gtk.Container] {
         hexpand: true
     })
 
+    const stack_tab_buttons_label = new Gtk.Label({
+        label: "Show Tab Buttons for Stacked Windows",
```
jacobgkau: Please rename the label here to `Show tabs for stacked windows` to match the gschema description.

jacobgkau (commented):
Looks almost good to go, enabling/disabling is behaving as expected now. I'll do full regression testing once the label's changed as previously discussed.

src/prefs.ts
awisus: @jacobgkau done

jacobgkau (approved):
The new toggle in the advanced extension settings works as intended and applies immediately when changed.

https://github.com/pop-os/shell/assets/7199422/cb97f077-c455-4088-a308-ef5a79813231

Regression testing (without the new hidden setting enabled) passed: [shell-pr1551-jammy.txt](https://github.com/pop-os/shell/files/13531891/shell-pr1551-jammy.txt)

mmstick (approved):

PR #1562, opened 2023-01-04
some targets should be run serially

xuhdev:
On some systems, "make" is configured to run parallelly by default. Here, debug and local-install should always be run serially.


mmstick (approved):

jacobgkau (approved):
Confirmed the multi-step targets currently fail from running in the wrong order when run in parallel with the `-j` option. Confirmed the `.NOTPARALLEL` option prevents the issue from occurring by forcing them to run in serial.

PR #1567, opened 2023-01-15
fix: allow search/launcher hotkeys to work while capslock is on

### LINKED ISSUE #102: Shortcuts Ctrl + J/K does not work when Caps Lock is ON
**Issue:** Shortcuts for navigating up/down in the launcher, Ctrl + J/K does not work

**pop-launcher version:**
```
pop-launcher:
  Installed: 1.2.1~1648660013~21.10~0893990
  Candidate: 1.2.1~1648660013~21.10~0893990
  Version table:
 *** 1.2.1~1648660013~21.10~0893990 1001
       1001 http://apt.pop-os.org/release impish/main amd64 Packages
        100 /var/lib/dpkg/status

```
**Pop OS:** 21.10 & now 22.04
**Other notes**: Other shortcuts like TAB & arrow keys seems to work as usual when Caps Lock is ON.

canadaduane:
Previously, only lower-case character codes were considered in hotkeys related to the search/launcher:

- Ctrl+k, Ctrl+p (up)
- Ctrl+j, Ctrl+n (down)
- Ctrl+q (quit)

This PR allows capslock to be on (or shift held down) while pressing hotkeys, and the same expected actions will occur.

Note that this also helps clean up the numeric constants used as key codes, and replaces them with easier-to-read key names.

Other hotkeys that may be affected (and would be worth testing) include:

- Ctrl+1 thru Ctrl+9
- Ctrl+Tab (autocomplete)
- Ctrl+Shift+Tab (up)
- Escape

Fixes https://github.com/pop-os/launcher/issues/102

### DIFF
--- a/src/search.ts
+++ b/src/search.ts
@@ -8,7 +8,7 @@ import type { JsonIPC } from 'launcher_service'
 
 const GLib: GLib = imports.gi.GLib;
 
-const { Clutter, Gio, Pango, Shell, St } = imports.gi;
+const { Clutter, Gio, Pango, Shell, St, Gdk } = imports.gi;
 const { ModalDialog } = imports.ui.modalDialog;
 
 const { overview, wm } = imports.ui.main;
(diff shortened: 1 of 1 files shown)


mmstick (approved):

n3m0-22 (approved):
This looks good to me the new behavior works as intended. I see no regression and all shell testing has passed. As an added benefit this also seems to allow `Ctrl` + `J/K` launcher navigation to work in Wayland as well.

[shell-testing.txt](https://github.com/pop-os/shell/files/10447909/shell-testing.txt)

PR #1573, opened 2023-01-23
fix: avoid activating a window that was/is being destroyed

### LINKED ISSUE #1572: Attempting to run a JS callback during garbage collection - forever spewing syslog
Hello
I've had a bug that  originally reported here. But **it only happens when extensions are enabled.**. so here I am.
https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/1868#note_1573153

This is the log I've prepared. When my python process creates a new window, bug happens after a while (sometimes it happens in a few seconds, sometimes takes 30s or more), then my gnome is completely unresponsive.
[log.txt](https://github.com/pop-os/cosmic/files/9852785/log.txt)

This log continues like this very long, and **it also fill many GB at syslog.**
I only use the default popOS extensions, and I haven't installed anything to gnome. Not even themes.

System:
`Linux pop-os 5.19.0-76051900-generic #202207312230~1663791054~22.04~28340d4 SMP PREEMPT_DYNAMIC Wed S x86_64 x86_64 x86_64 GNU/Linux`

mmstick:
@masip85 Check if this fixes the issue

Closes #1572

### DIFF
--- a/src/window.ts
+++ b/src/window.ts
@@ -127,7 +127,7 @@ export class ShellWindow {
     }
 
     actor_exists(): boolean {
-        return this.meta.get_compositor_private() !== null;
+        return this.meta.get_compositor_private() !== null || this.destroying;
     }
 
     private bind_window_events() {
@@ -663,6 +663,13 @@ export class ShellWindow {
 /// Activates a window, and moves the mouse point.
 export function activate(ext: Ext, move_mouse: boolean, win: Meta.Window) {
     try {
+        // Return if window was destroyed.
+        if (!win.get_compositor_private()) return
+
+        // Return if window is being destroyed.
+        if (ext.get_window(win)?.destroying) return
+
+        // Return if window has override-redirect set.
         if (win.is_override_redirect()) return
 
         const workspace = win.get_workspace()


13r0ck (approved):

jacobgkau (approved):
Regression testing passed: [shell-pr1537-jammy.txt](https://github.com/pop-os/shell/files/10494556/shell-pr1537-jammy.txt)

Arcitec:
@mmstick Hi, just checking if there's a typo in this part of the patch:

```js
    actor_exists(): boolean {
        return this.meta.get_compositor_private() !== null || this.destroying;
    }
```

So this thing says that the actor exists "if the actor has a non-null compositor assignment, or *if it LACKS a compositor assignment BUT the actor window is currently being destroyed*".

Is that intentional? Or should it have said the following instead:

```js
    actor_exists(): boolean {
        // compositor exists and window is not currently in the process of being destroyed
        return this.meta.get_compositor_private() !== null && !this.destroying;
    }
```

Logically the latter seems to make more sense, at a glance. But there could be some reason for the "weird" logic that I'm totally missing!

mmstick:
Yeah, should be !destroying.

Arcitec:
@mmstick Sweet, I see that the fix has been committed:

https://github.com/pop-os/shell/commit/a08804d2aca5371faad4c7f857ac49327ac445aa#diff-9e13cab61f6b4bbb06ef29422f0dc0251bcdfd15a9f139008b491f7d1a78e981R129-R131

I'm just curious about one thing to make sure: Is it proper to treat a window actor as "existing" even if it lacks a compositor? The `||` confuses me, since I had thought `&&` might make more sense there?

But I suppose that an actor **can** exist, **not** be in a destructing state, and **not** be assigned to any compositor, but **still** be a valid actor? In that case, yeah the final `||` makes sense there. :) :+1:

PR #1578, opened 2023-02-06
fix: actor_exists conditional flipped for destroying property

mmstick:
There's a typo in the conditional check. It should be `|| !this.destroying` so that an actor being destroyed is always assumed to no longer exist. The behavior here is virtually identical to what it was doing before, since the actor does exist when it's not yet destroyed.

### DIFF
--- a/src/window.ts
+++ b/src/window.ts
@@ -127,7 +127,7 @@ export class ShellWindow {
     }
 
     actor_exists(): boolean {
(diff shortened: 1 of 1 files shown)


13r0ck (approved):

XV-02 (approved):
This is a small alteration, and all the typical pop-shell features continue to work as expected, so calling it good.

PR #1579, opened 2023-02-08
Option to disable stacking with mouse

szmikler:
I am often stacking windows by mistake when dragging and dropping them using mouse. This PR adds an option to disable Mouse Stacking entirely. However, stacking is still available if activated explicitly with Super+S. One can then move windows between stacks using keyboard shortcuts instead of mouse.

image

I believe this might be the answer for the following issues:
* https://github.com/pop-os/shell/issues/965
* https://github.com/pop-os/shell/issues/1491

PR #1580, opened 2023-02-09
Extension breaks when stacked window is destroyed

mmstick:
Fixes a regression from the last commit which can cause windows in a stack being destroyed to break the extension. The conditional would consider an actorless window to have an actor if it's not yet marked as being destroyed. This'll correctly require that a window must have an actor and must not be set as destroying.

### DIFF
--- a/src/window.ts
+++ b/src/window.ts
@@ -127,7 +127,7 @@ export class ShellWindow {
     }
 
     actor_exists(): boolean {
-        return this.meta.get_compositor_private() !== null || !this.destroying;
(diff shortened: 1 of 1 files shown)


n3m0-22 (commented):
This fixes the issue in https://github.com/pop-os/shell/pull/1579. I will approve at the same time with https://github.com/pop-os/shell/pull/1579 as long as the rest of the shell testing passes.

n3m0-22 (approved):
Tested with #1579. All testing passed.

PR #1587, opened 2023-02-22
Allow user to Ctrl-C copy active option and close window

### LINKED ISSUE #173: [Feature Request] Copy active search result to clipboard on Ctrl + C
When using the launcher frontend, it is often convenient to be able to "copy" the active search result to the clipboard.

For example, when using the [calc](https://github.com/pop-os/launcher/blob/master/plugins/src/calc/mod.rs) math plugin, suppose the user enters a formula such as "1 + 2" into the frontend. A single calculated answer, "3", is shown and active in the launcher result list. If the user could press "Ctrl + C" at this point to copy the resulting number to the clipboard, it would accelerate workflows where the answer is an input to another app or website.

Implementation suggestions:

1. This "copy active result" action should only take place if there is no text highlighted in the frontend text input. In other words, if text is highlighted, the default "copy text to clipboard" system behavior should always take precedence. Only when there is nothing highlighted should "Ctrl + C" (or another copy shortcut key) mean "copy the highlighted answer to the clipboard".
2. A launcher result has two fields, `name` (shown as line 1) and `description` (shown as line 2). If the `description` is available, it should be preferred, as this is usually the "long form" of a result, such as the full path of a file. If `description` is null or blank, the `name` field should be used as a fallback (as in the `calc` plugin example above).
3. Some developers may not be aware that there are several "copy to clipboard" keys. If possible, all shortcuts should behave as expected. Here is a complete list of "copy to clipboard" keyboard shortcut keys:
- Ctrl + C
- Ctrl + Insert
- Copy [accelerator key](https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h#L211)

canadaduane:
Fixes https://github.com/pop-os/launcher/issues/173

### DIFF
--- a/src/launcher.ts
+++ b/src/launcher.ts
@@ -12,9 +12,12 @@ import type { Ext } from 'extension'
 import type { ShellWindow } from 'window'
 import type { JsonIPC } from 'launcher_service'
 
-const { Clutter, Gio, GLib, Meta, Shell } = imports.gi
+const { Clutter, Gio, GLib, Meta, Shell, St } = imports.gi
 
 const app_sys = Shell.AppSystem.get_default();
+
+const Clipboard = St.Clipboard.get_default();
+const CLIPBOARD_TYPE = St.ClipboardType.CLIPBOARD;
 
 interface SearchOption {
     result: JsonIPC.SearchResult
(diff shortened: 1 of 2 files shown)


mmstick (approved):

n3m0-22 (approved):
This is working as described in https://github.com/pop-os/launcher/issues/173#issue-1590482002.

Example:
Using `Ctrl` + `C` now copies the result of equations as well as the file location when using `find`. If the text in the input field of launcher is highlighted that text will be copied instead. This also works with other key combination like `Ctrl` + `Insert`. Other launcher plugins like `web` are still working as expected.

All testing passed.
[shell-1587-testing.txt](https://github.com/pop-os/shell/files/10808951/shell-1587-testing.txt)

Issue #1056, opened 2021-07-05
gjs file running


ZayRTun:
**(1) Issue/Bug Description:**
A file called `gjs` is always running but I can't switch to this application and I cannot quit it either.
Check the image file.
![20210705_095535](https://user-images.githubusercontent.com/31787822/124413858-7cd34780-dd77-11eb-8d34-03030ca9bce3.jpg)

**(2) Steps to reproduce (if you know):**
It simply appears after I boot the OS

**(3) Expected behavior:**
Shouldn't be there on it's own.

**(4) Distribution (run `cat /etc/os-release`):**
NAME="Pop!_OS"
VERSION="21.04"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 21.04"
VERSION_ID="21.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=hirsute
UBUNTU_CODENAME=hirsute
LOGO=distributor-logo-pop-os
 

**(5) Gnome Shell version:**
Gnome Shell version: 3.38.5

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**
```
pop-shell:
  Installed: 1.1.0~1625266132~21.04~78833ad
  Candidate: 1.1.0~1625433688~21.04~acd91df
  Version table:
     1.1.0~1625433688~21.04~acd91df 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu hirsute/main amd64 Packages
       1001 http://ppa.launchpad.net/system76/pop/ubuntu hirsute/main i386 Packages
 *** 1.1.0~1625266132~21.04~78833ad 100
        100 /var/lib/dpkg/status
```

**(7) Where was Pop Shell installed from:**
Was upgraded internally from pop os 20.04

**(8) Monitor Setup (2 x 1080p, 4K, Primary(Horizontal), Secondary(Vertical), etc):**
2 x 1080p

**(9) Other Installed/Enabled Extensions:**
User Themes

**(10) Other Notes:**

bflanagin:
[...]

PR #1605, opened 2023-04-02
move gjs from DEFAULT_FLOAT_RULES to SKIPTASKBAR_EXCEPTIONS

GloriousEggroll:
When pop-shell is installed alongside desktop-icons NG and "Show Minimize to Tray Windows" is enabled, gjs shows up as an application in the alt-tab menu (like in this previous bug: https://github.com/pop-os/shell/issues/1056). By completely moving it to the SKIPTASKBAR_EXCEPTIONS it gets resolved/no longer shows up when the option is enabled.


mmstick (approved):

n3m0-22 (approved):
I was able to reproduce the issue, and this does seem to fix it.

I found no regressions while testing. Looks good to me.

[shell-testing.txt](https://github.com/pop-os/shell/files/11152107/shell-testing.txt)

PR #1641, opened 2023-06-29
feat: allow active hint border width changes

### LINKED ISSUE #1631: Feature Request: Create a Setting to Change the Thickness of the Active Hint
Being able to change the thickness of active hint would be helpful because sometimes it is too thick and blocks information from adjacent screens. For example, the active hint with Discord app makes it hard to see the discord indicator for new posts.

shans10:
Closes #1631.

### DIFF
--- a/dark.css
+++ b/dark.css
@@ -10,7 +10,7 @@
 }
 
 .pop-shell-border-normal {
-    border-width: 3px;
+    border-width: var(--active-hint-border-width, 3px);
 }
 
 .pop-shell-border-maximize {
--- a/highcontrast.css
+++ b/highcontrast.css
@@ -10,7 +10,7 @@
 }
 
 .pop-shell-border-normal {
-    border-width: 3px;
+    border-width: var(--active-hint-border-width, 3px);
 }
 
 .pop-shell-border-maximize {
--- a/light.css
+++ b/light.css
@@ -10,7 +10,7 @@
 }
 
 .pop-shell-border-normal {
-    border-width: 3px;
+    border-width: var(--active-hint-border-width, 3px);
 }
 
 .pop-shell-border-maximize {
--- a/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
+++ b/schemas/org.gnome.shell.extensions.pop-shell.gschema.xml
@@ -13,6 +13,13 @@
             <summary>Border radius for active window hint, in pixels</summary>
         </key>
 
+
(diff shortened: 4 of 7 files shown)


mmstick (approved):

ghostdevv:
Hey, any chance we could merge this? :pray:

jacobgkau:
@pop-os/ux Can you confirm if this new option labeled "Active Border Width" is okay to be added to the top-right dropdown menu? I seem to recall asking about this before, but I can't find the messages now.

Screenshot from 2023-09-13 10-52-36

The new option works the same as the existing Active Border Radius option-- on current GNOME, it doesn't take effect immediately, but rather when focus goes back to the window.

Width seems like a more basic option than radius, so I think it makes sense to allow this. It might make sense to move width above radius in the menu.

jacobgkau:
@pop-os/ux Ping. Could you let us know if this option is okay to be added to 22.04, if it's not good for 22.04 but can be added to the branch for 23.10+ (which we're not shipping in Pop!_OS), or if it shouldn't be added at all?

maria-komarova:
Sorry for missing the ping earlier and thanks for pinging me now. We carefully selected the border width for active hint before. Introducing a setting for corner radius was a way to deal with the issue of differing corner radii for windows. If it would be possible to automatically detect it and set the corner radius for this, it would have been preferable to avoid having this setting. Overall I feel like we are overwhelming the tiling menu with more smaller customization options and that would be better to avoid unless the setting has a larger impact.

pyrrho:
My two cents as a Pop_OS! user, I'd love to see this change land. I think the defaults are very reasonable for users getting used to the system, but I wound up here specifically because I was looking for a way to lessen the visual weight of the active border, and disable it for all but the focused window.

I understand the concern about overpopulating the setting menu, though. My understanding is that one of the stated objectives of this project is to provide a light weight, easy to learn tiling window manager, and flooding users with configuration options is a great way to overload people and slow adoption. Were I designing this, I would hide the visual options behind a sub-menu, similar to the Floating Window Exceptions modal, or e.g. Slack's Troubleshooting in-menu accordion. This is very much my own personal preference, but providing an initially minimal surface area for configuration and interaction, while also leaving breadcrumbs to make the more advanced, hidden options discoverable for those willing to dig usually feels the most welcoming (again, to me).

McCio (approved):

nitin88:
@jacobgkau please merge?

PR #1654, opened 2023-09-24
improve(makefile): detect wayland on restart after build

pinsard:
Wayland doesn't support session restart, so a different approach has to be taken in order to prevent a blank screen that requires the user to start a new TTL and reboot (if she/he knows how to do it), or to force shutdown by holding down the power button.

Tested on Debian 12, 
Linux lt-5 6.1.0-12-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.52-1 (2023-09-07) x86_64 GNU/Linux

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,12 @@ $(info UUID is "$(UUID)")
 
 sources = src/*.ts *.css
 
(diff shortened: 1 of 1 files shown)


tmke8:
This seems to have been superseded by #1698.

PR #1657, opened 2023-10-03
Add "auto unstack" switch

MeowcaTheoRange:
Create a toggle switch in the settings to enable destroying stacks when they have only one window.

This is my first PR, and it "works on my machine" when I run `make debug`. There still might be bugs, though. 

My system specs:
```
OS: Nobara Linux 37 (Thirty Seven) x86_64 
Kernel: 6.3.5-201.fsync.fc37.x86_64 
Resolution: 1920x1080 
DE: GNOME 43.2 on Mutter X11
WM Theme: Adwaita
Theme: adw-gtk3-dark [GTK2/3] 
Icons: Adwaita [GTK2/3] 
CPU: Intel i7-10750H (12) @ 2.600GHz 
GPU: NVIDIA GeForce GTX 1660 Ti Mobile 
Memory: 7833MiB 
```


MeowcaTheoRange:
~~Oh, I should also explain the strange indent and .prettierignore file.
Well actually, I'm pretty sure they explain themselves if you've ever used Prettier.~~
Well, I fixed it, but Prettier also changed some other things. Will fix soon.

MeowcaTheoRange:
image
[ALT: A cake with the text "Sorry I committed code that was formatted by my IDE" "(it might happen again)"]

jacobgkau (changes requested):
This doesn't seem to be working. After turning the option on, I can still move windows out of a stack and have a single remaining window remain in the stack, rather than having the stack be destroyed.

https://github.com/pop-os/shell/assets/7199422/212efa32-4761-49eb-b139-3f66f7cb34e0

The most recent commit also broke the tab bar's formatting, as shown in the above video (the commit message includes `do not PR`, so I assume pushing it here was a mistake?)

MeowcaTheoRange:
> This doesn't seem to be working. After turning the option on, I can still move windows out of a stack and have a single remaining window remain in the stack, rather than having the stack be destroyed.
> simplescreenrecorder-2023-10-12_17.37.42.mp4
> 
> The most recent commit also broke the tab bar's formatting, as shown in the above video (the commit message includes `do not PR`, so I assume pushing it here was a mistake?)

Oh no, I didn't mean to PR the recent commit, I think it just did it automatically. I'll try to remove the commit.

Also, it might not be working because my commit might have only affected mouse stacking - I based my code off of the code that reverted this stacking behaviour, so I'm not sure how to also make it apply to keyboard stacking.

jacobgkau:
> Also, it might not be working because my commit might have only affected mouse stacking - I based my code off of the code that reverted this stacking behaviour, so I'm not sure how to also make it apply to keyboard stacking.

The current wording of the setting:

> Destroy a window stack when it contains only one window

...does not reflect that it's specific to the mouse. If you don't want to get it working for the keyboard, then you could alternatively tweak the wording so it's more indicative of what it does, e.g.:

> Destroy stacks after removing second-to-last window with mouse

(The wording could be nicer, but I tried to limit the length and also match the wording of existing settings in this panel.)

MeowcaTheoRange:
> > Also, it might not be working because my commit might have only affected mouse stacking - I based my code off of the code that reverted this stacking behaviour, so I'm not sure how to also make it apply to keyboard stacking.
> 
> The current wording of the setting:
> 
> > Destroy a window stack when it contains only one window
> 
> ...does not reflect that it's specific to the mouse. If you don't want to get it working for the keyboard, then you could alternatively tweak the wording so it's more indicative of what it does, e.g.:
> 
> > Destroy stacks after removing second-to-last window with mouse
> 
> (The wording could be nicer, but I tried to limit the length and also match the wording of existing settings in this panel.)

Well, I do want to accomodate for keyboard stacking, I just don't know how to.

But, if it's just fine with the mouse, it could be called "Destroy stacks when separated with mouse" or something like that.

src/prefs.ts
```diff
@@ -117,114 +133,123 @@ function settings_dialog_view(): [AppWidgets, Gtk.Container] {
...
 
     const smart_label = new Gtk.Label({
         label: "Smart Gaps",
-        xalign: 0.0
-    })
+        xalign: 0.0,
+    });
+
+    const unstack_label = new Gtk.Label({
+        label: "Automatically destroy single-window stacks",
```
jacobgkau: The label here still needs to be updated to match what https://github.com/pop-os/shell/pull/1657/commits/6adc8b28654737346aa1bec672883ee4c390b833 changed.

src/prefs.ts
```diff
@@ -117,114 +133,123 @@ function settings_dialog_view(): [AppWidgets, Gtk.Container] {
...
+        xalign: 0.0,
+    });
 
     const show_skip_taskbar_label = new Gtk.Label({
         label: "Show Minimize to Tray Windows",
-        xalign: 0.0
-    })
+        xalign: 0.0,
+    });
```
jacobgkau: It's probably best if you go ahead and revert the unnecessary formatting changes.

jacobgkau (changes requested):

src/prefs.ts
MeowcaTheoRange: Did not notice this, thanks!
image

src/prefs.ts
MeowcaTheoRange: Reverting these changes is how I found out VS Code has a shortcut specifically for saving without formatting. It turns out even without Prettier it likes to shuffle around the imports :/

src/auto_tiler.ts
MeowcaTheoRange, line 1: This is me trying to move the imports back to their rightful place

PR #1625, opened 2023-05-11
fix: correctly hide the decoration on gnome 44 (#1622)

kneirinck:
This PR fixes #1622 and is based on the suggestion https://github.com/velitasali/gtktitlebar/issues/36#issuecomment-1455278899

PR #1661, opened 2023-10-14
Fix the title bar showing/hiding

### LINKED ISSUE #1622: Title Bar doesn´t hide anymore on gnome 44
Title bar still visible with option show windows title off

toggle off windows itle in extension settings

**(3) Expected behavior:**

hide titles in gtk apps 

archlinux

**(5) Gnome Shell version:**

44.1

AUR

1080p

**(10) Other Notes:**

image

ghostdevv:
This is based off #1625 so Closes #1622 and #1625

I tested it on gnome 42 and 44 and it works on both, the only difference in the PRs is the change to the `may_decorate` functions. I'm not 100% sure what this function does but my best guess is to avoid adding `_MOTIF_WM_HINTS` if some other application has already done that. I adapted this check to include that the flags being set by pop shell are now `0x0` and `0x1`.

If this PR is good to go I'll also raise the exact same change for g45 on `master_mantic`

### DIFF
--- a/src/xprop.ts
+++ b/src/xprop.ts
@@ -77,7 +77,7 @@ export function get_xid(meta: Meta.Window): string | null {
 
 export function may_decorate(xid: string): boolean {
     const hints = motif_hints(xid);
-    return Array.isArray(hints);
+    return hints ? hints[2] == '0x0' || hints[2] == '0x1' : true;
 }
 
 export function motif_hints(xid: string): Array<string> | null {
--- a/src/xprop.ts
+++ b/src/xprop.ts
@@ -7,7 +7,7 @@ const GLib: GLib = imports.gi.GLib;
 const { spawn } = imports.misc.util;
 
 export var MOTIF_HINTS: string = '_MOTIF_WM_HINTS';
-export var HIDE_FLAGS: string[] = ['0x2', '0x0', '0x2', '0x0', '0x0'];
+export var HIDE_FLAGS: string[] = ['0x2', '0x0', '0x0', '0x0', '0x0'];
 export var SHOW_FLAGS: string[] = ['0x2', '0x0', '0x1', '0x0', '0x0'];
 
 export function get_window_role(xid: string): string | null {
@@ -77,7 +77,7 @@ export function get_xid(meta: Meta.Window): string | null {
 
 export function may_decorate(xid: string): boolean {
     const hints = motif_hints(xid);
-    return hints ? hints[2] != '0x0' : true;
+    return Array.isArray(hints);
 }
 
 export function motif_hints(xid: string): Array<string> | null {


Neo-29:
You're a life aver if it works
The  `GTK Title Bar` extension is broken and apparently noone seems to be on the move to update it for g45

ghostdevv:
Just want to ask @jacobgkau and @mmstick whenever you guys have spare time as you reviewed the last pr :pray:

jacobgkau (approved):
This gets the feature working on Fedora 38 with GNOME 44, and does not break the feature on Pop!_OS 22.04 with GNOME 42.

On Fedora 38, if tiling is turned off or a window is manually floated, I sometimes see the window have its size or position changed when turning titlebars back on (the size/position may be reverting to an earlier state somehow, based on my testing.) This did not happen on Fedora 38 with the current `master_jammy` branch. This doesn't happen in Pop!_OS 22.04, and it still seems like an improvement in behavior for Fedora 38.

Full regression testing passed:
[shell-pr1661-jammy.txt](https://github.com/pop-os/shell/files/13083930/shell-pr1661-jammy.txt)

Neo-29:
> This gets the feature working on Fedora 38 with GNOME 44, and does not break the feature on Pop!_OS 22.04 with GNOME 42.
> 
> On Fedora 38, if tiling is turned off or a window is manually floated, I sometimes see the window have its size or position changed when turning titlebars back on (the size/position may be reverting to an earlier state somehow, based on my testing.) This did not happen on Fedora 38 with the current `master_jammy` branch. This doesn't happen in Pop!_OS 22.04, and it still seems like an improvement in behavior for Fedora 38.
> 
> Full regression testing passed: [shell-pr1661-jammy.txt](https://github.com/pop-os/shell/files/13083930/shell-pr1661-jammy.txt)

does it work on G45?

mmstick (approved):

Neo-29:
Tks so much

Neo-29:
A quick question> When will it be in master mantic branch?
This is the one used on AUR
Tks

ghostdevv:
@Neo-29 Sent a pr #1669 to fix this

pbui:
I am experiencing a regression with this.  I have Zoom as a floating exception and with the old version, it looks like this:

Screenshot from 2023-10-30 15-46-13

With this PR, I now see a title bar:

Screenshot from 2023-10-30 15-47-05

I would note that Zoom is a Qt application, so I'm not sure if that makes a difference or not.

leviport:
@pbui which version of Gnome?

pbui:
@leviport This is with Gnome 42, running the latest Pop!_OS 22.04.

ghostdevv:
Testing with Zoom on both Gnome 42.5 and 45 shows that with this pr it doesn't hide the title bar. I don't know enough about how this works here, but when I was debugging this initially I could only find examples of using the set of flags this PR introduces not what we had previously.

I can raise another PR that only applies the different flag set to gnome 44 but it won't help fix it on 44 & 45, so any ideas there would be great

pbui:
I wonder if it is a Qt thing.  I just tested with OBS Studio and with this patch it shows the title bar, but without the patch it does not.

Update: I also tried with VLC (another Qt application) and the same thing.  It appears this PR breaks the feature with Qt applications (on Gnome 42.5).

Issue #1649, opened 2023-08-22
Gnome 45 support


Guara92:
**(1) Issue/Bug Description:**
Pop Shell is not compatible with Gnome 45, it's probably related to the ESM switch https://gjs.guide/extensions/upgrading/gnome-shell-45.html#esm 

```
import declarations may only appear at top level of a module
```

**(2) Steps to reproduce (if you know):**
Upgrade to Gnome 45

**(3) Expected behavior:**
should work

**(4) Distribution (run `cat /etc/os-release`):**
 ```
NAME="Fedora Linux"
VERSION="39 (Workstation Edition Prerelease)"
ID=fedora
VERSION_ID=39
VERSION_CODENAME=""
PLATFORM_ID="platform:f39"
PRETTY_NAME="Fedora Linux 39 (Workstation Edition Prerelease)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:39"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f39/system-administrators-guide/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=39
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=39
SUPPORT_END=2024-05-14
VARIANT="Workstation Edition"
VARIANT_ID=workstation
```

**(5) Gnome Shell version:**
45

**(6) Pop Shell version (run `apt policy pop-shell` or provide the latest commit if building locally):**
```gnome-shell-extension-pop-shell-1.2.0^11.b5accce-2.fc39.noarch```

**(7) Where was Pop Shell installed from:**
Fedora repo

carlwgeorge:
Howdy, maintainer for the Fedora package here.  Once this extension is ported to ESM (and the metadata.json files reflects GNOME 45 compatibility) I'll be happy to update the package accordingly.

RoyalOughtness:
[...]

PR #1676, opened 2023-11-14
fix: check if extension local schemas dir exists before use

lleyton:
Should resolve the issue the Fedora peeps are having in https://github.com/pop-os/shell/issues/1649 :3

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -32,15 +32,11 @@ function settings_new_id(schema_id: string): Settings | null {
 
 function settings_new_schema(schema: string): Settings {
     const GioSSS = Gio.SettingsSchemaSource;
-    // const schemaDir = Me.dir.get_child('schemas');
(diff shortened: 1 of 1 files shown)


carlwgeorge:
I included this change as a patch to the Fedora package, and it worked great.  Thanks!

s373r:
Just tested -- works great!

Many thanks to you!

mmstick (approved):

nitin88:
@jacobgkau Can you please review the same?

jacobgkau (approved):
The extension seems to work fine with this change on both Ubuntu 23.10 and Fedora 39. This does get rid of the error message in the logs when the `pop-shell@system76.com/schemas` dir isn't present on Fedora.

This change doesn't affect Pop!_OS since it's going into `master_mantic`.

PR #1734, opened 2024-10-01
fix: gnome 47 and meta 15 support

### LINKED ISSUE #1732: Gnome 47: TypeError when attempting to enable pop shell
Just upgraded to Gnome 47, Pop Shell encounters "TypeError: window.is_client_decorated is not a function" when I try to enable the extension from the extensions menu.

**(3) Expected behavior:**

Arch Linux

**(5) Gnome Shell version:**
Gnome 47

Commit [ced5442](https://github.com/pop-os/shell/commit/ced54427487905f42447a07dabff7101b700b5c9)

Git

1 x 2K

*[Arch Linux Updates Indicator](https://extensions.gnome.org/extension/1010/archlinux-updates-indicator/)
*[Auto Move Windows](https://extensions.gnome.org/extension/16/auto-move-windows/)
*[Blur My Shell](https://extensions.gnome.org/extension/3193/blur-my-shell/)
*[Caffeine](https://extensions.gnome.org/extension/517/caffeine/)
*[Dash to Dock](https://extensions.gnome.org/extension/307/dash-to-dock/)
*[Launch New Instance](https://extensions.gnome.org/extension/600/launch-new-instance/)
*[Native Window Placement](https://extensions.gnome.org/extension/18/native-window-placement/)
*[Removable Drive Menu](https://extensions.gnome.org/extension/7/removable-drive-menu/)
*[Rounded Window Corners Reborn](https://extensions.gnome.org/extension/7048/rounded-window-corners-reborn/) (slightly modified by me for an extra colour option)
*[Screenshot Window Sizer](https://extensions.gnome.org/extension/881/screenshot-window-sizer/)
*[Transparent Window Moving](https://extensions.gnome.org/extension/1446/transparent-window-moving/)
*[User Themes](https://extensions.gnome.org/extension/19/user-themes/)

I also tried disabling all of these and logging out and in, before attempting to start pop shell, same issue.

**(10) Other Notes:**

CodeAndGin:
Fix for #1732, updating window.is_client_decorated() to window.decorated

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1671,7 +1671,7 @@ export class Ext extends Ecs.System<ExtEvent> {
         }
 
         for (const window of this.windows.values()) {
-            if (window.meta.is_client_decorated()) continue;
+            if (window.meta.decorated) continue;
 
             if (show_title) {
                 window.decoration_show(this);
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1671,7 +1671,7 @@ export class Ext extends Ecs.System<ExtEvent> {
         }
 
         for (const window of this.windows.values()) {
-            if (window.meta.decorated) continue;
+            if (window.is_client_decorated()) continue;
 
             if (show_title) {
                 window.decoration_show(this);
--- a/src/mod.d.ts
+++ b/src/mod.d.ts
@@ -261,6 +261,7 @@ declare namespace Meta {
         appears_focused: Readonly<boolean>;
         minimized: Readonly<boolean>;
         window_type: Readonly<any>;
+        decorated: Readonly<boolean>;
 
         activate(time: number): void;
         change_workspace_by_index(workspace: number, append: boolean): void;
@@ -285,7 +286,6 @@ declare namespace Meta {
         has_focus(): boolean;
         is_above(): boolean;
         is_attached_dialog(): boolean;
-        is_client_decorated(): boolean;
         is_fullscreen(): boolean;
         is_on_all_workspaces(): boolean;
         is_override_redirect(): boolean;
--- a/src/window.ts
+++ b/src/window.ts
@@ -63,7 +63,7 @@ export class ShellWindow {
     ignore_detach: boolean = false;
     was_attached_to?: [Entity, boolean | number];
     destroying: boolean = false;
-
+    
     // Awaiting reassignment after a display update
     reassignment: boolean = false;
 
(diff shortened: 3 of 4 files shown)


jackpot51 (dismissed):

jacobgkau (changes requested):
Thank you for the update!

Testing on Fedora 40 with GNOME 46, this breaks the `Show Window Titles` toggle, which is available under Xorg. As you can see in the below screenshot, LibreOffice still has its title even when the option's turned off, and the terminal (with CSD) appears inside of an extra frame (this happens if the window's open before I switch the option off).

Screenshot from 2024-10-03 17-21-10

Neither of these things happen with the current `master_noble` branch, as you can see below. (If I recreate the above issue on the new version, then install the old version and restart GNOME Shell, the terminal still has the frame until I close it and re-open it, but I can't replicate it from scratch on the old version.)

Screenshot from 2024-10-03 17-23-16

If possible, can you gate this behavior change behind a check of the GNOME Shell version? You can see an example of doing that here: https://github.com/pop-os/shell/blob/ced54427487905f42447a07dabff7101b700b5c9/src/extension.ts#L2036 

If that's not possible, then we'll need to create another branch for GNOME 47+. Our branches are integrated with our CI, which usually tracks Ubuntu releases; it does look like Oracular (24.10) has GNOME 47, so we could make a `master_oracular` if necessary (although I don't think our CI is set up for it yet).

CodeAndGin:
Been looking at this in a bit more depth. Found a version of an older Meta api. window.is_client_decorated() and window.decorated are not, as I naively assumed, cognate. is_client_decorated() is checking if the decorations are being handled by the window, and it's used to continue the for loop because otherwise the weird frame thing happens.

is_client_decorated() was removed from the API, which is _annoying_ because stuff like this is outside of my expertise and I don't know how to hack together something that can check for what is_client_decorated() was checking for.

Basically, whoops this wasn't a fix at all because I missed the _why_ and a simple version check won't solve the issue except on 46 and earlier.

CodeAndGin:
I've just pushed a new commit; I _think_ the new function I wrote is doing the same hacky thing that the API version of is_client_decorated() was doing, based on the comments in the old code that defined it. I am _not_ particularly versed in how xprop properties work though, so I'm kinda running blind here. _However_ what I've written _seems_ to do the expected behaviour correctly on my system. If it still needs a version gate let me know.

jacobgkau (approved):
On 209ab01, there no longer seems to be any regression on GNOME 46. The window titles option works normally under Xorg. That option is not available on Wayland, but tiling is still working as expected on Wayland.

McCio (approved):
LGTM

mmstick (approved):

PR #1756, opened 2024-12-07
fix: "Tile Windows" toggle causing recursion when using shortcut

### LINKED ISSUE #1733: Desktop freezes when disabling tiling mode via shortcut

Entire desktop freezes (except for cursor) after disabling tiling mode via shortcut (Super + Y). The shortcut works normally when enabling tiling mode (after disabling it via the menu).

1. Clean install of EndeavourOS.
2. Build Pop Shell from source or install gnome-shell-extension-pop-shell-git from AUR.
3. Press Super + Y (other Pop Shell shortcuts work)
4. Entire desktop freezes (except cursor)

**(3) Expected behavior:**

Tiling shortcut should not cause the desktop to freeze and should disable tiling mode the same way as from the menu.

```
NAME="EndeavourOS"
PRETTY_NAME="EndeavourOS"
ID="endeavouros"
ID_LIKE="arch"
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://endeavouros.com"
DOCUMENTATION_URL="https://discovery.endeavouros.com"
SUPPORT_URL="https://forum.endeavouros.com"
BUG_REPORT_URL="https://forum.endeavouros.com/c/general-system/endeavouros-installation"
PRIVACY_POLICY_URL="https://endeavouros.com/privacy-policy-2"
LOGO="endeavouros"
```

**(5) Gnome Shell version:**

47

Local build commit: ced54427487905f42447a07dabff7101b700b5c9
AUR package: gnome-shell-extension-pop-shell-git 1.2.0.r305.gcfa0c55-1

Manual install from source or gnome-shell-extension-pop-shell-git from AUR.

3440 x 1440 horizontal primary, 1440 x 2560 vertical secondary

None

**(10) Other Notes:**

Toggling tiling mode via the menu works as expected.

While the desktop seems to freeze completely, the cursor is still able to move around. CTRL + ALT + F3 does not switch terminal. 

The only way to recover from the freeze is via force shutdown (power button hold).

Remarkable is that the CPU fans will ramp up when the freeze happens and do not slow down.

Hardware: AMD Ryzen 7 5800X and Radeon RX 6800

epetousis:
Fixes #1733, it was an infinite recursion issue [that was caused by updating the state for the "Tile Windows" toggle in the indicator popup.](https://github.com/pop-os/shell/issues/1733#issuecomment-2524836497)

I fixed this by calling the appropriate `auto_tile_*` method inside the `connect` callback for the toggle. If this isn't the preferred way to go about it, then I'm happy to change it.

Works on my machine, even if I press and hold the `toggle-tiling` shortcut key combination.

### DIFF
--- a/src/panel_settings.ts
+++ b/src/panel_settings.ts
@@ -308,13 +308,13 @@ function show_title(ext: Ext): any {
     return t;
 }
 
-function toggle(desc: string, active: boolean, connect: (toggle: any) => void): any {
+function toggle(desc: string, active: boolean, connect: (toggle: any, state: boolean) => void): any {
     let toggle = new PopupSwitchMenuItem(desc, active);
 
     toggle.label.set_y_align(Clutter.ActorAlign.CENTER);
 
-    toggle.connect('toggled', () => {
-        connect(toggle);
+    toggle.connect('toggled', (_: any, state: boolean) => {
+        connect(toggle, state);
         return true;
     });
 
@@ -322,7 +322,13 @@ function toggle(desc: string, active: boolean, connect: (toggle: any) => void): 
 }
 
 function tiled(ext: Ext): any {
-    let t = toggle(_('Tile Windows'), null != ext.auto_tiler, () => ext.toggle_tiling());
+    let t = toggle(_('Tile Windows'), null != ext.auto_tiler, (_, shouldTile) => {
+        if (shouldTile) {
+            ext.auto_tile_on();
+        } else {
+            ext.auto_tile_off();
+        }
+    });
     return t;
 }
 


jaxwilko:
Tested and fixes the issue on Debian trixie/sid Gnome 47.2 :+1:

manavchaudhary1:
Seems to be bugs on Fedora 41 Gnome 47.2
https://imgur.com/a/5SyFXDB
Opens another instance.

jaxwilko:
Running `make local-install` calls `restart-shell` which can be found here: https://github.com/pop-os/shell/blob/e25621e2595eb5235ecb1a41167d1324a2b2a297/Makefile#L70

Could it be causing the odd behaviour for you?

On Nvidia my desktop always crashes if I run `make local-install` from within the DE, so I switch to a different TTY (i.e. CTRL + ALT + 6) then run the command in the tty without a DE. Switching back to Gnome after seems to work fine after doing this.

manavchaudhary1:
Now it's working after installing from tty3 and rebooting.

manavchaudhary1:
Ctrl + Enter not coloring window blue , which causes confusion.
Problem specific to this pr working fine in main branch.

epetousis:
> Ctrl + Enter not coloring window blue , which causes confusion.

This sounds like you have my `main` branch checked out, which is where I'm tinkering with the extension and aiming to implement moving windows without having to hit Ctrl + Enter. Make sure you have the commit epetousis/pop-shell@af83e5e checked out.

mmstick (approved):

jacobgkau (approved):
Just to keep this straight in case it's important later, the bug in question affects GNOME 47 but not GNOME 46 and below. Ubuntu Noble (24.04) comes with GNOME 46, and we don't have an Oracular branch yet. However, this should be fine, because the fix doesn't seem to cause any regressions on GNOME 46. Toggling the state on and off with both the top-right icon and the keyboard shortcut (and a combination of the two) works normally with this branch on Pop!_OS 24.04 Alpha, as it did before.

On GNOME 47 (testing with EndeavourOS), I was able to recreate the issue before this fix (by simply toggling tiling off with the keyboard shortcut and observing a freeze), and am no longer seeing it occur after the fix.

This is not merging into a stable Pop!_OS branch and does not affect the currently supported 22.04 session.

Thank you for digging into this and contributing the fix!

dagimg-dot:
Tested on Fedora 41 Gnome 47.2 and the issue is fixed. Thank you all!

PR #1772, opened 2025-03-09
metadata: Add GNOME 48.

siddhpant:
Pre-requisite: #1771.


mmstick (approved):

jacobgkau (approved):
Tested on Fedora 42 Alpha with GNOME 48. This PR allows the extension to be enabled, and basic functionality works when this is applied on top of https://github.com/pop-os/shell/pull/1771 (which has already been merged).

This is not merging into a branch being used within stable Pop!_OS.

PR #1777, opened 2025-04-14
fix: keyboard shortcuts for left/right workspaces

### LINKED ISSUE #1776: Workspace switching shortcuts & workspace menu don't work
The shortcuts `Ctrl+Super+<Direction>` and `Super+D` for workspace navigation don't work on my system (= nothing happens when I press those key combinations). Everything else that I tried (tiling, switching windows, keyboard shortcuts for maximizing/minimizing/toggle float, window stacking) works as expected. Correct dconf entries exist:

```
$ dconf read /org/gnome/desktop/wm/keybindings/switch-to-workspace-down
['<Primary><Super>Down', '<Primary><Super>j']
$ dconf read /org/gnome/desktop/wm/keybindings/switch-to-workspace-up
['<Primary><Super>Up', '<Primary><Super>k']
```

Maybe this has something to do with switching workspaces left/right vs up/down? Touchpad gestures for switching workspaces left/right (three-finger swipe) work fine.

## Context

- System: Debian 12
- GNOME Version 43
- Running on Wayland
- Installed using `make local-install`, from newest commit `0f7a5d1` on branch `master_jammy`

Thanks for developing this project!

laikq:
Add keyboard shortcuts for left/right workspace switching & window moving.

- `SUPER + Ctrl + left/right` for workspace switching
- `SUPER + Ctrl + Shift + left/right` for moving windows between workspaces

Fix #1776 

Hopefully useful for some people installing pop-shell on vanilla gnome.

### DIFF
--- a/scripts/configure.sh
+++ b/scripts/configure.sh
@@ -47,9 +47,9 @@ set_keybindings() {
     # Show the activities overview: disable <Super>s
     dconf write ${KEYS_GNOME_SHELL}/toggle-overview "@as []"
     # Switch to workspace left: disable <Super>Left
-    dconf write ${KEYS_GNOME_WM}/switch-to-workspace-left "@as []"
+    dconf write ${KEYS_GNOME_WM}/switch-to-workspace-left "@as ['<Primary><Super>Left','<Primary><Super>${left}']"
     # Switch to workspace right: disable <Super>Right
-    dconf write ${KEYS_GNOME_WM}/switch-to-workspace-right "@as []"
+    dconf write ${KEYS_GNOME_WM}/switch-to-workspace-right "@as ['<Primary><Super>Right','<Primary><Super>${right}']"
     # Maximize window: disable <Super>Up
     dconf write ${KEYS_GNOME_WM}/maximize "@as []"
     # Restore window: disable <Super>Down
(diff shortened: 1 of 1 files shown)


jacobgkau (changes requested):
Because this script isn't part of the actual extension source code or package, this won't affect typical Pop!_OS 22.04 usage with the packaged version of pop-shell. It will only affect manual installations on other distros, or on Pop!_OS for the purpose of development.

On Debian 12 with GNOME 43, I confirmed this makes the Ctrl-Super-Left/Right shortcuts switch horizontal workspaces as expected.

On Pop!_OS 22.04, the Ctrl-Super-Left/Right shortcuts snap a window to the left or right side of the screen by default. I installed the packaged version of this to double-check that no changes were observed. I then installed using `make local-install` to have the script applied to the system; after that, the Ctrl-Super-Left/Right shortcuts didn't do anything, because they'd been assigned to switch horizontal workspaces, but Pop!_OS 22.04 doesn't use horizontal workspaces.

With that in mind, I don't think this should be merged as-is, since it breaks one of the expected shortcuts on a Pop!_OS system. On Debian, the left/right snap shortcut is just Super-Left/Right by default, which Pop!_Shell obviously already overrides with its shortcuts to move focus to the left/right, which is why this regression wouldn't be noticed on Debian.

There are a couple of paths forward for this PR:

- Apply instead to `master_noble`. Although this won't help with Debian 12 since it uses GNOME 43, this extension is no longer being used in Pop!_OS 24.04 (Noble), nor is the vertical workspaces extension; that means Pop!_OS's vertical workspaces are no longer a concern, nor is breaking Pop!_OS compatibility. The change could be made on that branch to benefit future users on later versions of GNOME.
- A check could be made either for the OS (don't apply the left/right workspace shortcuts on Pop!_OS) or for the existing shortcuts (don't apply the left/right workspace shortcuts if those shortcuts are assigned to snap windows) in the configure script.

jacobgkau:
If we did want to apply this to `master_noble`, we'd want to give some thought to whether a different pair of shortcuts could take on the role of snapping windows to the left/right. Although I agree navigating workspaces is probably more useful than snapping as a keyboard shortcut (especially since snapping happens outside of the tiling workflow), snapping has long been part of the Pop!_Shell regression testing checklist and is thus a feature we intended to support with the Pop!_Shell experience. (The simplest solution would be swapping it to Ctrl-Super-Up/Down since those wouldn't be in use with horizontal workspaces, but there'd be no intuitive answer to whether Up should snap left or right, and vice versa.)

PR #1786, opened 2025-08-21
Edited Inactive Tab Color for More Intuitive Selection

andrewtran3643:
The inactive tab color was hard-coded to `#9B8E8A`. When using a dark active hint color, the inactive tab appears lighter than the active tab, which can make it seem like the inactive tab is selected.

This PR updates the inactive tab color to use the active hint color, darkened by 20%, and slightly adjusts the text color to be darker for inactive tabs.

This is my first PR—thank you for your patience, and I appreciate any feedback!

image

### DIFF
--- a/src/stack.ts
+++ b/src/stack.ts
@@ -14,7 +14,7 @@ import St from 'gi://St';
 const ACTIVE_TAB = 'pop-shell-tab pop-shell-tab-active';
 const INACTIVE_TAB = 'pop-shell-tab pop-shell-tab-inactive';
 const URGENT_TAB = 'pop-shell-tab pop-shell-tab-urgent';
-const INACTIVE_TAB_STYLE = '#9B8E8A';
+const INACTIVE_DARKEN_AMOUNT = -0.2;
 
 export var TAB_HEIGHT: number = 24;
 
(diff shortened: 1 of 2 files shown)


andrewtran3643:
Hi everyone! Just checking in—has there been any interest or feedback on this PR? Are any adjustments needed?

Also, here's what it looks like in another color with transparency:
image

jacobgkau:
@andrewtran3643 Hadn't gotten a chance to look at this yet, sorry (our attention's mostly turned towards [COSMIC Epoch](https://github.com/pop-os/cosmic-epoch), which no longer uses Pop!_Shell). I'll take a look at this on behalf of QA tomorrow. I do think it makes sense to derive the inactive tab color from the active hint color so it matches the active tab color. Shouldn't be any problems as long as it works as expected.

jacobgkau (commented):
Before:
Screenshot From 2025-08-28 19-26-57
Screenshot From 2025-08-28 19-27-09
Screenshot From 2025-08-28 19-27-29

After:
Screenshot From 2025-08-28 21-42-26
Screenshot From 2025-08-28 21-41-00
Screenshot From 2025-08-28 21-40-32

This is definitely an improvement for non-default active hint colors. However, I'm a little concerned that the difference between active & inactive tabs is less noticeable now.

@andrewtran3643 Would you be able to try and change the shading to match the inactive tab color to what it used to be a little closer when the hint color's set to the default? It doesn't have to be 100% exact, but it seems like e.g. reducing the saturation a bit might get closer to what it used to be. Seems like it could also benefit darker hint colors.

Issue #1793, opened 2025-09-22
Gnome 49 support


CodeAndGin:
_Well my Gnome 47 fix had a good run but bye bye x11 I guess_

While the extension does load when adding "49" to the metadata, toggling tiling does nothing.

Neo-29:
+1 ( archlinux wayland here)

rafafrdz:
+1 (cachyos wayland here) added "49" and pluggin is enabled but toggling tiling doesn't work

LilMonk:
> +1 (cachyos wayland here) added "49" and pluggin is enabled but toggling tiling doesn't work

Same for me

rr987-0:
i've made several function modifications get_maximized to is_maximized and some others but can exactly remember what but here is a version that works. https://drive.google.com/file/d/1g_SzZyO6HeDX0hS7fhQTzeEsDUVaT98r/view?usp=sharing 
just unzip it and replace  .local/share/gnome-shell/extensions/pop-shell@system76.com 
or diff it and figure out

Neo-29:
> i've made several function modifications get_maximized to is_maximized and some others but can exactly remember what but here is a version that works. https://drive.google.com/file/d/1g_SzZyO6HeDX0hS7fhQTzeEsDUVaT98r/view?usp=sharing just unzip it and replace .local/share/gnome-shell/extensions/[pop-shell@system76.com](mailto:pop-shell@system76.com) or diff it and figure out

Why not a PR?

rafafrdz:
> i've made several function modifications get_maximized to is_maximized and some others but can exactly remember what but here is a version that works. https://drive.google.com/file/d/1g_SzZyO6HeDX0hS7fhQTzeEsDUVaT98r/view?usp=sharing 
> just unzip it and replace  .local/share/gnome-shell/extensions/pop-shell@system76.com 
> or diff it and figure out

yes please, could you open a Pr with the changes? @rr987-0

LilMonk:
@rafafrdz @CodeAndGin @Neo-29 You can check out this PR https://github.com/pop-os/shell/pull/1795 . Let me know if there are any bugs or issues with it. Its working fine for me on cachyos. `Note: Need to reboot after installation.`

chrisjonesBSU:
[...]

PR #1795, opened 2025-09-24
Fix: Patch for gnome 49

LilMonk:
This patch addresses the migration to gnome 49 #1793


GGisMee:
Worked well when tested!

jacobgkau (approved):
Confirmed on EndeavourOS that this allows enabling Pop!_Shell on GNOME 49, and it appears to function normally.

Confirmed on Pop!_OS 24.04 with GNOME installed that this still functions on GNOME 46.

Maximization and attempting to perform tiling operations while a window's maximized seem to work as expected on both. Mouse follows cursor doesn't work under Wayland, but already didn't before.

mmstick (approved):

PR #1809, opened 2026-03-16
Fix tiled windows positioned behind top bar on workspace move

### LINKED ISSUE #1565: Windows are partly covered behind top bar
Hi everyone!
I spend some time searching for this issue in this repository but could not find something that matched my problem.
Thank you all so much for your consideration! 

With tiling on, windows are resized to the complete height of the screen, so that the top of the window is covered by the top bar. 
Some workspaces are fine, while some have this behavior. Here is an example with firefox: 
Workspace 10_001

It happens often, but I cannot predict when. It seems to happen more often when moving windows to another workspace.

**(3) Expected behavior:**
Window is only as high as the space below the top bar, so that all elements can be clicked on.

```
 NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
VERSION_CODENAME=jammy
UBUNTU_CODENAME=jammy
LOGO=distributor-logo-pop-os
```

**(5) Gnome Shell version:**
```
GNOME Shell 42.3.1
```

```
pop-shell:
  Installed: 1.1.0~1667482449~22.04~30bf682
  Candidate: 1.1.0~1667482449~22.04~30bf682
  Version table:
 *** 1.1.0~1667482449~22.04~30bf682 1001
       1001 http://apt.pop-os.org/release jammy/main amd64 Packages
       1001 http://apt.pop-os.org/release jammy/main i386 Packages
        100 /var/lib/dpkg/status
```

OS installation (Pop-OS)?

1x4K, horizontal, scale 100% without fractional scaling. HiDPI Daemon disabled (did not affect behavior when enabled and is laggy)

User Themes 49
Sound Input & Output Device Chooser 43

**(10) Other Notes:**
Could this be similar as  #1232?

cagostino:
monitor_work_area() always queries the active workspace for work area geometry, even when tiling a window that just moved to a different workspace. when a window gets moved between workspaces, the workspace-changed signal fires and pop-shell detaches + reattaches the window to the new workspace's tiling tree. during that reattach, it calls monitor_work_area() to figure out where to place the window — but since the active workspace hasn't switched yet, get_work_area_for_monitor() returns geometry for the wrong workspace. depending on timing, this can return y=0 instead of y=panel_height, so the window gets tiled behind the top bar.

the fix adds an optional workspace_idx parameter to monitor_work_area() so callers that know which workspace they're targeting can pass it explicitly instead of relying on the active workspace. all tiling codepaths that have fork/workspace context (attach_to_monitor, update_toplevel, detach_window, smart gap recalculations) now pass the actual workspace index. callers that genuinely operate on the active workspace (launcher, snap, display readiness check) are left unchanged.

this also explains why the bug is intermittent — it depends on whether gnome has finished switching the active workspace by the time the tiling code runs.

fixes #1565

### DIFF
--- a/src/auto_tiler.ts
+++ b/src/auto_tiler.ts
@@ -95,7 +95,7 @@ export class AutoTiler {
     }
 
     update_toplevel(ext: Ext, fork: Fork, monitor: number, smart_gaps: boolean) {
-        let rect = ext.monitor_work_area(monitor);
+        let rect = ext.monitor_work_area(monitor, fork.workspace);
 
         fork.smart_gapped = smart_gaps && fork.right === null;
 
@@ -120,7 +120,7 @@ export class AutoTiler {
 
     /** Attaches `win` to an optionally-given monitor */
     attach_to_monitor(ext: Ext, win: ShellWindow, workspace_id: [number, number], smart_gaps: boolean) {
-        let rect = ext.monitor_work_area(workspace_id[0]);
+        let rect = ext.monitor_work_area(workspace_id[0], workspace_id[1]);
 
         if (!smart_gaps) {
             rect.x += ext.gap_outer;
@@ -153,7 +153,7 @@ export class AutoTiler {
             if (monitor) {
                 if (fork.is_toplevel && fork.smart_gapped && fork.right) {
                     fork.smart_gapped = false;
-                    let rect = ext.monitor_work_area(fork.monitor);
+                    let rect = ext.monitor_work_area(fork.monitor, fork.workspace);
 
                     rect.x += ext.gap_outer;
                     rect.y += ext.gap_outer;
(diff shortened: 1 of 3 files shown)


mmstick (approved):

jacobktm:
I haven't been able to recreate this issue before applying this PR in order to verify that this resolves the issue. I noticed that the OS in the referenced issue is pop 22.04 but this PR looks like it's targeting pop 24.04, are you also having this issue in pop 24.04 and if so is there a set of steps I could use to recreate it?

jacobgkau:
@jacobktm It sounds like the issue's intermittent (sounds like a race condition). We can probably just regression test this since engineering approved the approach.

The PR targeting the Noble branch means this is for users of the extension on non-Pop distros (or would apply if someone's replaced COSMIC with GNOME on Pop).

cagostino:
actually no it should be targeting 22.04, i don't know if this issue occurs on 24.04 so that is my bad for specifying that incorrectly.

jacobgkau:
Ok. @cagostino Your earlier PR, https://github.com/pop-os/shell/pull/1808, was targeting the correct branch for 22.04 (`master_jammy`), but included a lot of unnecessary (and breaking) commits since it was based on the default `master_noble` branch. Can you rebase on `master_jammy`?

cagostino:
yeah i'll get this worked out tn

cagostino:
Superseded by #1812, which is rebased on master_jammy.

PR #1811, opened 2026-03-27
Add support for Gnome 50

BernatBC:
Adds support for gnome 50.

Environment:
- Ubuntu 26.04 beta
- gnome-shell/resolute,now 50.0-0ubuntu2
- Typescript version 6.0.2

Screenshot:
image


jacobgkau (approved):
This is still working as expected with GNOME 46 on Ubuntu 24.04, both on Wayland and X11, and including floating exceptions (looking at what was touched in the code).

The extension did not previously work with GNOME 50 on Ubuntu 26.04 Beta, but does after these changes.

This branch is not currently used in the default Pop!_OS environment.

mmstick (approved):

Issue #1758, opened 2024-12-23
Border hint isn't shown for first window even after tiling


siddhpant:
Border hint isn't shown for first maximised window.

That's nice, but when I open something else and it tiles, the hint won't show up for the first original window, but it will show up for new windows.

If I close the original window and restart the program, I do have border for both the tiles. That is to say, the border stays for the new window even after the original one closed and the new one got maximized / is the only one.

siddhpant:
PS: #1703 is a different issue, it's a feature request rather than a bug report.

Issue #1820, opened 2026-06-17
Keybind overwrite spam when entering/exiting window management mode


siddhpant:
> Also there is spam of the following when exiting edit mode with Esc:
> 
> ```
> Overwriting existing binding of keysym 73 with keysym 73 (keycode 27).
> Overwriting existing binding of keysym ff54 with keysym ff54 (keycode 74).
> Overwriting existing binding of keysym ff52 with keysym ff52 (keycode 6f).
> Overwriting existing binding of keysym ff54 with keysym ff54 (keycode 74).
> Overwriting existing binding of keysym ff51 with keysym ff51 (keycode 71).
> Overwriting existing binding of keysym ff51 with keysym ff51 (keycode 71).
> Overwriting existing binding of keysym ff53 with keysym ff53 (keycode 72).
> Overwriting existing binding of keysym 68 with keysym 68 (keycode 2b).
> Overwriting existing binding of keysym 6c with keysym 6c (keycode 2e).
> Overwriting existing binding of keysym ff52 with keysym ff52 (keycode 6f).
> Overwriting existing binding of keysym ff53 with keysym ff53 (keycode 72).
> ``` 

 _Originally posted by @siddhpant in [#1801](https://github.com/pop-os/shell/issues/1801#issuecomment-3500975418)_

Issue #1819, opened 2026-06-13
Terminal windows momentarily shrink and break tiling layout during clipboard operations


Argus-Khan:
## Bug Description

Whenever `wl-clipboard` performs a clipboard operation, the active terminal window briefly shrinks for a split second. This occasionally corrupts the tiling layout.

This did not always happen and appears to have started recently.

## Steps to Reproduce

1. Enable tiling mode
2. Open any terminal with a tiled layout
3. Run any of the following:
   - `wl-copy` or `wl-paste` directly in the terminal
   - Copy a file path in **Yazi** (file manager)
   - Yank text in **Neovim**
4. Observe the terminal window briefly shrink and occasionally snap into a broken tiling layout

## Expected Behavior

Clipboard operations should have no effect on window geometry or tiling layout.

## Actual Behavior

The active terminal window shrinks momentarily during any `wl-clipboard` operation. This occasionally causes the tiling layout to break and requires manual correction.

https://github.com/user-attachments/assets/8f35fcd7-2d86-46a2-95be-e6c3680f1dbb

## Additional Notes

- Tested across **three different terminals**: Kitty, Foot, and WezTerm — issue occurs in all of them
- Tested with `wl-copy --foreground` — makes no difference
- Affects both `wl-copy` and `wl-paste`
- Also triggered by any application that uses `wl-clipboard` internally (confirmed with Yazi and Neovim)
- Compiled pop-shell from the latest git repo to check if it was already fixed — issue persists
- Screen recording attached

## Environment

| | |
|---|---|
| **OS** | Fedora Linux 44 (Workstation Edition) x86_64 |
| **Host** | Dell OptiPlex 3070 |
| **Kernel** | 7.0.11-200.fc44.x86_64 |
| **GNOME** | 50.2 |
| **WM** | Mutter (Wayland) |
| **pop-shell version** | 2 (also tested latest git build) |
| **wl-clipboard version** | 2.2.1 |
| **Terminals tested** | Kitty 0.47.1, Foot, WezTerm |
| **GPU** | Intel UHD Graphics 630 (Integrated) |
| **Display** | Dual monitor — 2x 2560x1440 @ 60Hz |

Argus-Khan:
[...]

Issue #1801, opened 2025-11-07
Stacked mode causes gnome-shell warnings about assertion errors


siddhpant:
```
Ignoring length property that isn't a number at line 4, col 20
meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
JS WARNING: file:///home/siddh/.local/share/gnome-shell/extensions/pop-shell@system76.com/window.js:264:22: Too many arguments to method Meta.Window.unmaximize: expected 0, got 1
JS WARNING: file:///home/siddh/.local/share/gnome-shell/extensions/pop-shell@system76.com/extension.js:640:31: Too many arguments to method Meta.Window.unmaximize: expected 0, got 1
```

Repro steps:

1. Enable stacked mode.
2. Open a new window. I used `gedit` for above.

siddhpant:
Also there is spam of the following when exiting edit mode with Esc:

```
Overwriting existing binding of keysym 73 with keysym 73 (keycode 27).
Overwriting existing binding of keysym ff54 with keysym ff54 (keycode 74).
Overwriting existing binding of keysym ff52 with keysym ff52 (keycode 6f).
Overwriting existing binding of keysym ff54 with keysym ff54 (keycode 74).
Overwriting existing binding of keysym ff51 with keysym ff51 (keycode 71).
Overwriting existing binding of keysym ff51 with keysym ff51 (keycode 71).
Overwriting existing binding of keysym ff53 with keysym ff53 (keycode 72).
Overwriting existing binding of keysym 68 with keysym 68 (keycode 2b).
Overwriting existing binding of keysym 6c with keysym 6c (keycode 2e).
Overwriting existing binding of keysym ff52 with keysym ff52 (keycode 6f).
Overwriting existing binding of keysym ff53 with keysym ff53 (keycode 72).
```

PR #1821, opened 2026-06-17
Some fixes for long-standing issues

siddhpant:


siddhpant:
Closes: #1758 #1794 #1819 #1820 #1801

README.md
```diff
@@ -68,6 +68,7 @@ Use the branch corresponding to your GNOME Shell version (`git checkout branch_n
 - **GNOME 42 through 44:** Use the `master_jammy` branch.
 - **GNOME 45:** Use the `master_mantic` branch.
 - **GNOME 46+:** Use the `master_noble` branch.
+- **GNOME 49+:** Use the `master_resolute` branch.
```
jacobgkau, line 71: Does this mean this branch is not backwards compatible back to 46+? We'll need to set a new branch up before merging if that's the case. (And the previous line should be updated to say `GNOME 46 through 48`.)

siddhpant: It's due to the maximise API change at https://gitlab.gnome.org/GNOME/mutter/-/merge_requests/4415.

I think the current branch is also broken for same reason, but I didn't test building on old GNOME.

siddhpant: To be clear, the other bug fixes does pertain to older versions too.

If we want a linear split in history, I can reorder commits from the current `master_noble` branch.

siddhpant: (Updated the README as requested)

siddhpant: If reordering doesn't sound good, new branch can start from `d22d2f` (=> master_noble at `d22d2f~1`). Only 3 commits.

And then the fixes can be back-ported (cherry-picked).

siddhpant: @jacobgkau ping

siddhpant: Ping? @jacobgkau

jacobgkau: > To be clear, the other bug fixes does pertain to older versions too.

It sounds like they should've been in a separate PR, in that case. One PR with all of that stuff to `master_noble`, then another PR that includes the 49+ fix for the new `master_resolute` branch. Does that make sense?

siddhpant: It would be the same history in the end, so if commits lgty then we can probably skip the intermediate step.

siddhpant: Basically the split has already happened implicitly due to `d22d2ff05989`, which has broken `master_noble` for GNOME < 49.

siddhpant: Actually let me make a new branch for `master_noble` having the revert and open a PR.

siddhpant: Ok done, created a new PR for master_noble, and this branch builds over it.

PR #1827, opened 2026-08-19
Master noble fixed

siddhpant:


siddhpant:
@jacobgkau for master-noble. the other branch is now rebased over it.

jacobgkau:
Reference: https://github.com/pop-os/shell/pull/1821

jacobgkau (changes requested):
Attempting to test this on GNOME 46 (on Pop!_OS 24.04 with GNOME Shell installed), I'm running into an issue where the window management shortcut keys (e.g. `s`, `o`, etc.) get stuck performing window management actions even when window management mode isn't active. This prevents me from typing those letters into any app windows (although they still work in the Activities overview). This doesn't seem to happen on the current `master_noble`. Perhaps it's a problem with f9576b8?

siddhpant:
Do you see any error in logs? I tried in a VM and it doesn't happen.

https://github.com/user-attachments/assets/f60a3f8a-5289-4ccb-8b0a-0dfa8e9e800c

siddhpant:
Ping

PR #1830, opened 2026-09-10
Add support for Gnome 51

BernatBC:
Adds support for gnome 51.

This upgrade was strightforward. Updating `metadata.json` was enough.

Environment:
- Ubuntu 26.10 Stonking Stingray (development branch) - Snapshot 3
- gnome-shell/stonking,now 51~beta-0ubuntu1
- Typescript version 6.0.3

Screenshot:
Screenshot From 2026-09-10 14-49-07


adil192 (changes requested):
This works on 51.beta but not on the newer 51.rc.

For 51.rc support, you need to replace St.BoxLayout's `vertical` property with `orientation` (see https://gjs.guide/extensions/upgrading/gnome-shell-51.html). I did this in my fork with https://github.com/adil192/shatter-shell/commit/b01d46508fc65183a407affce983ed43ff818520.

Note that `orientation` is only supported on GNOME 48 and newer.

PR #1832, opened 2026-09-17
fix: various small fixes

adil192:
Some cherrypicked fixes from my Shatter Shell fork, https://github.com/pop-os/shell/issues/1833.

***

- [X] I have disclosed use of any AI generated code in my commit messages.
  - If you are using an LLM, and do not fully understand the changes it is making to the code base, do not create a PR.
  - In our experience, AI generated code often results in overly complex code that lacks enough context for a proper fix or feature inclusion. This results in considerably longer code reviews. Due to this, AI authored or partially authored PRs may be closed without comment.
- [X] I understand these changes in full and will be able to respond to review comments.
- [X] My change is accurately described in the commit message.
- [X] My contribution is tested and working as described.
- [X] I have read the [Developer Certificate of Origin](https://developercertificate.org/) and certify my contribution under its conditions.


Voyagerroc-Lab (approved):
Reviewed with assistance from OpenAI Codex. The fixes correct entity comparison, storage growth bounds, slot-zero handling, and storage removal when the index is zero. Each change addresses a concrete boundary case in the affected logic.

pamod-madubashana (commented):
Bugfix review:

Summary: Cherry-picked small fixes. Real bugs fixed: `entity_eq` compared `b[1]==b[1]` (always true), `Storage.insert` inverted length check, `unregister_storage` treated index 0 as falsy, `slot` 0 treated as empty, `pid -1 vs 0` check. All correct.

1. `entity_eq`: `==` -> `===` good, fixes generation mismatch that could alias entities.
2. `insert`: `if(length>=id) fill` -> `if(id>=length) fill` correct. Confirm `fill(null,length,id)` end-exclusive is intended.
3. New float rules (`firefox PiP`, `git-credential-manager`, `xwaylandvideobridge`) look safe, but `git-credential-manager` float could hide auth prompt behind tiling - verify UX.
4. `TAB_HEIGHT` refactor changes `dpi*24` to `dpi*TAB_HEIGHT` - confirm constant equals 24 or call out visual change.

Please squash to conventional commits. LGTM otherwise.

adil192:
> Confirm fill(null,length,id) end-exclusive is intended.

Yes, no need to fill in the `id`th index to null when we set it in the next line.

But you did make me look over the [`Array.prototype.fill`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill) function and it turns out it does nothing if we don't increase the length first. Since javascript is so loosey-goosey, it makes no real difference at runtime (blank places were `undefined` instead of `null`) but it's fixed now.

> look safe, but git-credential-manager float could hide auth prompt behind tiling - verify UX.

No issues here, the `git-credential-manager` popup shows above other windows.

> `TAB_HEIGHT` - confirm constant equals 24 or call out visual change

`TAB_HEIGHT` does equal 24, no visual change.

> Please squash to conventional commits. LGTM otherwise.

They're already conventional commits. Is there something specifically which needs changing @pamod-madubashana?

